PR #29, opened 2022-03-07
Add simulateLeft and simulateRight to Simulate class

PietPtr:
`simulateRight` and `simulateLeft` can be used to simulate circuits that have the protocol `()` on one side. Such circuits are a useful model of [Manager/Subordinate](https://en.wikipedia.org/wiki/Master/slave_(technology)) circuits. 

This pull-request does not provide implementations of `simulateRight` and `simulateLeft` for every protocol included in clash-protocols, only for `()` and `Df`.


PietPtr:
We decided to go with a different (more general) way of simulating circuits that covers this case. See also pull request #32.

PR #30, opened 2022-03-22
Setup nix

basile-henry:


.github/workflows/ci.yml
```diff
@@ -101,6 +101,14 @@ jobs:
         run: |
           .ci/build_docs.sh
 
+  nix:
+    name: Nix build
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - uses: cachix/install-nix-action@v16
```
basile-henry, line 109: This is going to need a bit more setup: https://nix.dev/tutorials/continuous-integration-github-actions

martijnbastiaan: Cool. I think I've got time to look into it tomorrow.

basile-henry: Well I think it should work as is, it'll just be really slow until we setup cachix in particular because it builds `clash` from source.

Cachix should be free though since this is an open source/public project 😊

PR #31, opened 2022-04-01
Remove injectivity requirement of Fwd

PietPtr:
Since some protocols must be able to send `()` (or some other very simple type) over the `Fwd` channel, `Fwd` cannot be injective (as the protocol `()` already sends `()` over its `Fwd` channel).

One example of such a `Protocol` is a protocol such as `CSignal`, but in the other direction. Its definition:

```haskell
instance Protocol (CSignalR dom a) where
  type Fwd (CSignalR dom a) = CSignalR dom ()
  type Bwd (CSignalR dom a) = CSignalR dom a
```

This protocol cannot exist when `Fwd` is required to be injective.

Removing injectivity makes type errors much less informative. Sadly `Fwd` simply is not injective so we cannot benefit from this feature.

### DIFF
--- a/src/Protocols/Internal.hs
+++ b/src/Protocols/Internal.hs
@@ -161,7 +161,7 @@ instance Default a => Default (CSignal dom a) where
 class Protocol a where
   -- | Sender to receiver type family. See 'Circuit' for an explanation on the
   -- existence of 'Fwd'.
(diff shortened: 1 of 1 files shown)


martijnbastiaan (approved):
Sad, but necessary.

PR #39, opened 2022-06-17
Add datatypes and instances for AXI4-Stream protocol

samalws:
This PR adds datatypes and instances (Protocol, Backpressure, Simulate, Drivable, DfLike, Test) for the AXI4-Stream protocol (https://developer.arm.com/documentation/ihi0051/a/Introduction/About-the-AXI4-Stream-protocol).

I'm unsure if the way I implemented DfLike is the right way to go; I wanted the data sent over the channel to be of the form `Vec n Axi4StreamByte`, but DfLike requires any type to be allowed, so I used a type variable `dataType` with the added constraint that `dataType ~ Vec n Axi4StreamByte`.

### DIFF
--- a/clash-protocols.cabal
+++ b/clash-protocols.cabal
@@ -149,7 +149,7 @@ library
     Protocols.Axi4.WriteAddress
     Protocols.Axi4.WriteData
     Protocols.Axi4.WriteResponse
-    Protocols.Axi4.Stream.Axi4Stream
+    Protocols.Axi4.Stream
 
     Protocols.Df
     Protocols.DfConv
--- a/clash-protocols.cabal
+++ b/clash-protocols.cabal
@@ -128,6 +128,7 @@ library
     , deepseq
     , extra
     , ghc >= 8.6
+    , hashable
     , hedgehog >= 1.0.2
     , pretty-show
     , strict-tuple
@@ -148,6 +149,7 @@ library
     Protocols.Axi4.WriteAddress
     Protocols.Axi4.WriteData
     Protocols.Axi4.WriteResponse
+    Protocols.Axi4.Stream.Axi4Stream
 
     Protocols.Df
     Protocols.DfConv
--- a/src/Protocols/Axi4/Stream/Axi4Stream.hs
+++ b/src/Protocols/Axi4/Stream/Axi4Stream.hs
@@ -58,28 +58,20 @@ type family DestWidth (conf :: Axi4StreamConfig) where
 -- | Shorthand for a "well-behaved" config, so that we don't need to write out
 -- a bunch of type constraints later. Holds for every configuration; don't worry
 -- about implementing this class.
-type GoodAxi4StreamConfig conf =
+type KnownAxi4StreamConfig conf =
   ( KnownNat (DataWidth conf)
   , KnownNat (IdWidth conf)
   , KnownNat (DestWidth conf)
   )
 
--- | A byte sent along an AXI4 Stream. Each byte can either be a data byte, a
--- position byte, or a null byte. The value of position and null bytes should
--- be ignored. Additionally, null bytes can be added or dropped.
-data Axi4StreamByte
-  = DataByte (Unsigned 8)
-  | PositionByte
-  | NullByte
-  deriving (Generic, C.NFDataX, C.ShowX, Eq, NFData, Show, Hashable)
-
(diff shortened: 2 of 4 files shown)


src/Protocols/Axi4/Stream/Axi4Stream.hs
```diff
@@ -0,0 +1,233 @@
...
+  | NullByte
+  deriving (Generic, C.NFDataX, C.ShowX, Eq, NFData, Show, Hashable)
+
+-- | Data sent from manager to subordinate. The tvalid field is left out: messages with @tvalid = False@
+-- should be sent as a @NoAxi4StreamM2S@. The tdata, @tstrb@, and @tkeep@ fields are
+-- all grouped in the @_tdata@ field in this datatype (see @Axi4StreamByte@).
+data Axi4StreamM2S (conf :: Axi4StreamConfig) (userType :: Type)
+  = NoAxi4StreamM2S
+  | Axi4StreamM2S
+    { _tdata :: Vec (DataWidth conf) Axi4StreamByte
```
martijnbastiaan, line 82: This is definitely the Clashy way to define the spec, but I'm wondering if we really want this. There are two considerations:

 * How this is represented in HDL? I'm assuming having this at a top entity does not give you an API compatible AXI4 port mapping.
 * If we don't care about the port mapping: wouldn't it be Classier to simply make that `a`?

It seems like the current implementation gives us neither the flexibility we'd really want, nor API compatible HDL. I think I'd prefer to define `_tdata :: a`, with a separate fields for `_tkeep`/`_tstrb`. To create an API compatible component we'd simply need to make sure that `a ~ BitVector n` and the `_tkeep`/`_strb` are set accordingly.

Alternatively, we could define `_tdata :: BitVector n` and assume that most/all Clash designs don't want to deal with AXI4 and would therefore convert them to `Df` at their borders.

What do you think?

martijnbastiaan (changes requested):
LGTM! Let's resolve the discussions posted in-line before merging.

src/Protocols/Axi4/Stream/Axi4Stream.hs
```diff
@@ -0,0 +1,233 @@
...
+  type Bwd (Axi4Stream dom conf userType) = Signal dom Axi4StreamS2M
+
+instance Backpressure (Axi4Stream dom conf userType) where
+  boolsToBwd _ = C.fromList_lazy . fmap Axi4StreamS2M
+
+instance (GoodAxi4StreamConfig conf, NFDataX userType) =>
+  DfConv.DfConv    (Axi4Stream dom conf userType) where
+  type Dom         (Axi4Stream dom conf userType) = dom
+  type FwdPayload  (Axi4Stream dom conf userType)
+    = (Axi4StreamExtraInfo conf userType, Vec (DataWidth conf) Axi4StreamByte)
```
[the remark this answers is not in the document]
samalws: Good point, there wasn't really any reason to keep them separated

src/Protocols/Axi4/Stream/Axi4Stream.hs
samalws: Good point, I'll add `_tkeep` and `_tstrb` separately

samalws:
At this point there's essentially no difference between `Maybe Axi4StreamInfo` and `Axi4StreamM2S`, and I think dropping `Axi4StreamM2S` and replacing it everywhere with `Maybe Axi4StreamInfo` (or `Data Axi4StreamInfo`) would make things easier overall. What do you think @martijnbastiaan ?

martijnbastiaan:
Oh right, using `Maybe` is a very obvious way of handling this of course! Clash can even pry of the constructor bit at top entities and turn it into a `valid` port. Let's do that.

martijnbastiaan:
If that's done this PR can be merged :+1:

clash-protocols.cabal
```diff
@@ -148,6 +149,7 @@ library
     Protocols.Axi4.WriteAddress
     Protocols.Axi4.WriteData
     Protocols.Axi4.WriteResponse
+    Protocols.Axi4.Stream.Axi4Stream
```
martijnbastiaan, line 152: How come this is called `Stream.Axi4Stream`? Wouldn't `Stream` suffice, given that it is already under `Protocols.Axi4`?

martijnbastiaan (approved):

PR #43, opened 2022-07-08
Alternative DFLike class

samalws:
This PR adds an alternative, more general, DFLike class, which for now is called `DfLikeAlternative`. The class supports taking user parameters, as well as keeping state. Protocols implementing this class must create a `Circuit` with `Df` on one end and the protocol on the other. The PR also adds a few helper functions, instances, and unit tests.

Some questions / things to note:
- Should we remote the original `DfLike` class or just add this class in addition to it?
- In this PR I implemented `DfLikeAlternative` for `Df` and for `Axi4`. In another branch [here](https://github.com/samalws/clash-protocols/tree/dflike-alternative), I implemented and tested it for a bunch of other protocols (`Axi4Stream`, `AvalonStream`, and `AvalonMemMap`); I didn't add them here since support these protocols hasn't been added yet to the `main` branch.
- I can't test my `Axi4` instances without `instance Test (Axi4 ...)`, so I'm not sure that they're correct.
- I still need to write `Axi4` instances for master ports and for extra features e.g. burst modes.
- I've ported over a few `Df` functions, such as `map` and `zipWith`, to the new class in order to give some examples of how easily it can be done. I'm not sure what should be done with the full range of `Df` functions; should we port them all over, causing there to be a bunch of boilerplate code in `DfLikeAlternative.hs`, or leave them in `Df.hs`, forcing users to do a little extra work to use them with other protocols? Alternatively, we could put them all in `DfLikeAlternative.hs` and take them out of `Df.hs`, since there is an instance of `DfLikeAlternative` for `Df` anyway.


src/Protocols/DfLikeAlternative.hs
```diff
@@ -0,0 +1,416 @@
...
+  type Dom df :: Domain
+  -- | Information being sent in
+  type InpPayload df
+  -- | Information being sent out
+  type OtpPayload df
+  -- | User-provided parameters (e.g. address to respond to)
+  type DfLikeParam df
+  -- | Circuit which converts Df into this protocol's messages.
+  -- There are two Df channels, one for read data and one for write data
+  dfLikeCkt :: (HiddenClockResetEnable (Dom df))
```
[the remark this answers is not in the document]
martijnbastiaan: Do we get `fromDfCircuit` for free using `Reverse`? If so, I'd add a function doing this instead of requiring the user to fiddle with `Reverse`.

src/Protocols/DfLikeAlternative.hs
```diff
@@ -0,0 +1,416 @@
...
+    => DfLikeCktArgs df
+    -> Circuit (Df (Dom df) (OtpPayload df), Reverse (Df (Dom df) (InpPayload df))) df
+
+-- | Arguments given to 'dfLikeCkt'
+type DfLikeCktArgs df = (Proxy df, DfLikeParam df)
```
martijnbastiaan, line 51: Just inline this definition, it's only two arguments which makes me think the benefit of the extra level of indirection isn't justified by its costs.

src/Protocols/DfLikeAlternative.hs
```diff
@@ -0,0 +1,416 @@
...
+            put (Nothing,
+                 if _wlast then S2M_WriteResponse {_bid = sr, _bresp = ROkay, _buser = wrUser } else respS2M)
+            pure (S2M_WriteData{_wready = True}, Just _wdata)
+   
+      processWRespAck = when (_bready wRespAck) $ modify (\(a,_) -> (a,S2M_NoWriteResponse))
+
+-- Always sends 'ROkay' along 'WriteResponse' channel
+instance (NFDataX dat, NFDataX rdUser, KnownNat (Width aw), KnownNat (Width iw)) =>
+  DfLikeAlternative
+    (Reverse (Axi4ReadAddress dom 'KeepBurst 'NoSize lw iw aw keepRegion 'KeepBurstLength keepLock keepCache keepPermissions keepQos raData),
```
martijnbastiaan, line 173: I think I've asked this before, but I forgot: why are some of these properties hardcoded, and some aren't?

We should probably move this to a single type argument like you did in another PR, btw.

martijnbastiaan (changes requested):
In general, I really like this. It gets much closer to what I intended to do with `DfLike`, but never really managed. 

> Should we remote the original DfLike class or just add this class in addition to it?

Remove the original one.

> I've ported over a few `Df` functions, such as `map` and `zipWith`, to the new class in order to give some examples of how easily it can be done. I'm not sure what should be done with the full range of `Df` functions; should we port them all over, causing there to be a bunch of boilerplate code in `DfLikeAlternative.hs`, or leave them in `Df.hs`, forcing users to do a little extra work to use them with other protocols? Alternatively, we could put them all in `DfLikeAlternative.hs` and take them out of `Df.hs`, since there is an instance of `DfLikeAlternative` for `Df` anyway.

Almost everything in `Df.hs` refers to `DfLike.hs`, which is a pattern we should keep IMO. This:

```haskell
-- | Like 'P.fst', but over payload of a Df stream.
fst :: Circuit (Df dom (a, b)) (Df dom a)
fst = DfLike.fst
```

is much easier to read than its generalized (`DfLike`) counterpart. We can still do this if we replace `DfLike` by `DfLikeAlternative`, right?

All other questions: makes sense.

A few general comments:

 * We've got a style guide over here: https://github.com/clash-lang/clash-compiler/blob/master/STYLE.rst. Like the style guide itself says: almost all "rules" in there are strong suggestions.
 * Try to add more documentation and see how it renders in Haddock. `clash-protocols` is fairly complicated, so it can do with a lot of docs.

src/Protocols/DfLikeAlternative.hs
```diff
@@ -0,0 +1,416 @@
...
+-- Supports both input data and output data
+class (Protocol df) => DfLikeAlternative df where
+  -- | Domain that messages are being sent over
+  type Dom df :: Domain
+  -- | Information being sent in
+  type InpPayload df
+  -- | Information being sent out
+  type OtpPayload df
+  -- | User-provided parameters (e.g. address to respond to)
+  type DfLikeParam df
```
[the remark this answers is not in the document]
samalws: I updated the comment; better now? The gist is that we want to have addresses be parameterized so that circuits can respond to different addresses. For protocols without addresses (or anything else that needs to be parameterized), DfLikeParam can be set to ()

src/Protocols/DfLikeAlternative.hs
samalws: The hardcoded properties are the ones that `dfLikeCircuit` looks at, since there's currently no support for "looking, but with a default value if the field isn't there". That other PR adds support for looking with a default value, so once that's added we can get rid of the hardcoded properties.

samalws:
Rebased and added some changes. Copy of the old version is at [dflike-alternative-2-backup](https://github.com/samalws/clash-protocols/tree/dflike-alternative-2-backup)

src/Protocols/DfLike.hs
```diff
@@ -41,931 +68,1546 @@ module Protocols.DfLike
...
+-- so that they can be dealt with uniformly using 'Df'.
+-- For pipelined protocols, which can carry both in the same cycle,
+-- the 'Circuit' should pass along the interesting parts
+-- but not the uninteresting parts.
+--
+-- Can take parameters, e.g. for addresses.
+-- Defaults to being the right side of the circuit (@Fwd ~ Otp, Bwd ~ Inp@),
+-- but this can be switched using 'Reverse' from 'Protocols.Internal'.
+-- Supports both bwd (input/read) data and fwd (output/write) data.
+class (Protocol df) => DfLike df where
```
martijnbastiaan, line 124: Perhaps we should rename this `ToDf` or `DfConv`? It's not really that the protocols have to be anything like `Df`, the only requirement is that they can be converted to and from it.

src/Protocols/DfLike.hs
```diff
@@ -41,931 +68,1546 @@ module Protocols.DfLike
...
+  -- such as addresses, bursts, pipelining, etc. so that
+  -- a circuit connected to the 'Df' end doesn't have to worry about all that.
+  -- There are two Df channels, one for fwd data and one for bwd data,
+  -- so data can be sent both ways at once.
+  -- This circuit is expected to follow all of the conventions of 'Df';
+  -- for example, 'Df.Data' should stay the same
+  -- between clock cycles unless acknowledged.
+  toDfCircuit :: (HiddenClockResetEnable (Dom df))
+    => (Proxy df, DfLikeParam df)
+    -> Circuit (Df (Dom df) (FwdPayload df), Reverse (Df (Dom df) (BwdPayload df))) df
```
martijnbastiaan, line 157:
```suggestion
  toDfCircuit :: 
    (HiddenClockResetEnable (Dom df)) =>
    (Proxy df, DfLikeParam df) ->
    Circuit (Df (Dom df) (FwdPayload df), Reverse (Df (Dom df) (BwdPayload df))) df
```

main motivation for this style is that we'll get even more type symbols if we ever start using linear types. It also makes copy+pasting a bit better, as [explained here](https://yairchu.github.io/posts/silly-haskell-formatting).

src/Protocols/DfLike.hs
```diff
@@ -41,931 +68,1546 @@ module Protocols.DfLike
...
+  type BwdPayload df = ()
+  type FwdPayload df = ()
+  type DfLikeParam df = ()
+
+-- | Class for protocols whose 'DfLike' implementation contains
+-- a lot of info (e.g. addresses, user channels) that typically isn't needed.
+-- Using 'ImplicitInfo' we can put in default values for extraneous info
+-- and only deal with the important data (see 'ImplicitInfo' and instances below).
+class (DfLike df) => ImplicitInfoClass df where
+  -- | The "important" parts of @BwdPayload df@
```
martijnbastiaan, line 169: This should move to a separate module. It probably also should move to a different PR; I think we should wait and evaluate `DfLike` in practice before adding this.

src/Protocols/DfLike.hs
```diff
@@ -41,931 +68,1546 @@ module Protocols.DfLike
...
+    f ~(a, _) c = ((c, P.pure NoData), a)
+
+
+-- DfLike classes for Axi4
+
+-- | Data carried along 'Axi4WriteAddress' channel which is put in control of the user,
+-- rather than being managed by the 'DfLike' instances.
+-- Matches up one-to-one with the fields of 'M2S_WriteAddress' except for
+-- '_awlen', '_awsize', and '_awburst'.
+-- TODO where should I put this and 'Axi4ReadAddressInfo'?
```
martijnbastiaan, line 280: Move it to `ReadAddress.hs`.

src/Protocols/DfLike.hs
```diff
@@ -41,931 +68,1546 @@ module Protocols.DfLike
...
+          P.pure
+            ( S2M_ReadData
+              { _rid = readId
+              , _rlast = burstLenLeft == 1
+              , _rdata, _ruser, _rresp }
+            , _rready dataAck)
+        _ -> P.pure (S2M_NoReadData, False)
+
+instance
+  ( GoodAxi4ReadAddressConfig confAR
```
martijnbastiaan, line 612: Bit late, but I think we should rename this to `Axi4ReadAddressConfigConstraints`. That can go in a different PR.

src/Protocols/DfLike.hs
```diff
@@ -41,931 +68,1546 @@ module Protocols.DfLike
...
+
+-- 'fromDfCircuit', but on a pair of 'DfLike's
+tupFromDfLike
+  :: DfLike dfA
+  => DfLike dfB
+  => Dom dfA ~ Dom dfB
+  => HiddenClockResetEnable (Dom dfA)
+  => ((Proxy dfA, DfLikeParam dfA), (Proxy dfB, DfLikeParam dfB))
+  -> Circuit (Reverse (dfA, dfB))
+             ((Df (Dom dfA) (BwdPayload dfA), Df (Dom dfB) (BwdPayload dfB)), (Reverse (Df (Dom dfA) (FwdPayload dfA)), Reverse (Df (Dom dfB) (FwdPayload dfB))))
```
martijnbastiaan, line 852:
```suggestion
tupFromDfLike ::
  ( DfLike dfA
  , DfLike dfB
  , Dom dfA ~ Dom dfB
  , HiddenClockResetEnable (Dom dfA) ) =>
  ( (Proxy dfA, DfLikeParam dfA)
  , (Proxy dfB, DfLikeParam dfB) ) ->
  Circuit 
    (Reverse (dfA, dfB))
    ( ( Df (Dom dfA) (BwdPayload dfA)
      , Df (Dom dfB) (BwdPayload dfB) )
    , ( Reverse (Df (Dom dfA) (FwdPayload dfA))
      , Reverse (Df (Dom dfB) (FwdPayload dfB)) )
    )
```

A bit of formatting makes it much easier to read what goes into what tuple

src/Protocols/DfLike.hs
```diff
@@ -1,27 +1,54 @@
...
+for it and subsequently expose all the functionality implemented in this module.
 -}
+
 {-# LANGUAGE FlexibleContexts #-}
 {-# LANGUAGE FlexibleInstances #-}
-{-# LANGUAGE FunctionalDependencies #-}
-{-# LANGUAGE TypeFamilyDependencies #-}
+{-# LANGUAGE MultiParamTypeClasses #-}
 {-# LANGUAGE NamedFieldPuns #-}
+{-# LANGUAGE UndecidableInstances #-}
```
martijnbastiaan, line 11: `UndecidableInstances` can lead to infinite loops at compile time. It's fine to use it, but you need to explain why you need it.

clash-protocols.cabal
```diff
@@ -148,6 +150,7 @@ library
     Protocols.Axi4.WriteResponse
 
     Protocols.Df
+    -- Protocols.DfLike
```
martijnbastiaan, line 153:
```suggestion
```

src/Protocols/DfLike.hs
```diff
@@ -41,931 +68,1546 @@ module Protocols.DfLike
...
+
+-- | Class for protocols that are "similar" to 'Df',
+-- i.e. they can be converted into a 'Df' port using a 'Circuit' (see 'toDfCircuit').
+-- This is for protocols that carry some "interesting" data,
+-- as well as some "uninteresting" data (e.g. address, burst length).
+-- The 'Circuit' should abstract away the complexities of each protocol,
+-- so that they can be dealt with uniformly using 'Df'.
+-- For pipelined protocols, which can carry both in the same cycle,
+-- the 'Circuit' should pass along the interesting parts
+-- but not the uninteresting parts.
```
martijnbastiaan, line 118:
```suggestion
-- | Class for protocols that are "similar" to 'Df', i.e. they can be converted
-- into a 'Df' port using a 'Circuit' (see 'toDfCircuit'). This is for protocols
-- that carry some "interesting" data, as well as some "uninteresting" data
-- (e.g. address, burst length). The 'Circuit' should abstract away the
-- complexities of each protocol, so that they can be dealt with uniformly using
-- 'Df'. For pipelined protocols, which can carry both in the same cycle, the
-- 'Circuit' should pass along the interesting parts but not the uninteresting
-- parts.
```

You don't have to split up each sentence. Write documentation like you'd write a book :).

martijnbastiaan: A bit tedious to fix, but this isn't the only case that'd need fixing.

src/Protocols/DfLike.hs
martijnbastiaan: This would need to be applied in other places in this PR too.

martijnbastiaan (changes requested):

src/Protocols/DfLike.hs
```diff
@@ -41,931 +68,1546 @@ module Protocols.DfLike
...
+  -- This is the information being carried over the protocol,
+  -- /not/ the messages being carried over the protocol,
+  -- so it doesn't include auxiliary information like address or burst length.
+  -- If no data is sent in this direction, set this to @()@.
+  type FwdPayload df
+  -- | User-provided parameters for 'toDfCircuit'
+  -- (e.g. address to respond to, so that different 'DfLike'
+  -- components can respond to different addresses).
+  -- If you don't need to take params, set this to @()@.
+  type DfLikeParam df
```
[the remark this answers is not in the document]
samalws: `DfLikeParam` made a lot more sense before adding the address info etc to `FwdPayload` and `BwdPayload`, now the user can just explicitly see addresses. I think we should remove it.

src/Protocols/Df.hs
```diff
@@ -413,54 +536,217 @@ roundrobin ::
...
+                              else (brRead, amtLeft)
+        -- adjust blockram read address and amount left
+        (rAddr', amtLeft'') = if (amtLeft_ < maxBound && popped)
+                              then (C.satSucc C.SatWrap rAddr, amtLeft'+1)
+                              else (rAddr, amtLeft')
+        brReadAddr = rAddr'
+        -- return our new state and outputs
+        otpAck = Maybe.isJust maybePush
+        otpDat = if amtLeft_ < maxBound then Data brRead_ else NoData
+    in  ((rAddr', wAddr', amtLeft''), (brReadAddr, brWrite, Ack otpAck, otpDat))
```
martijnbastiaan, line 691:
```suggestion
    in  ((rAddr1, wAddr1, amtLeft2), (brReadAddr, brWrite, Ack otpAck, otpDat))
```

It's easy to miss a `'` when counting, so replace them by numbers. If you're using `rAddr1`, `rAddr0` should also exist.

src/Protocols/Df.hs
```diff
@@ -413,54 +536,217 @@ roundrobin ::
...
+                              then (Maybe.fromJust maybePush, amtLeft')
+                              else (brRead, amtLeft)
+        -- adjust blockram read address and amount left
+        (rAddr', amtLeft'') = if (amtLeft_ < maxBound && popped)
+                              then (C.satSucc C.SatWrap rAddr, amtLeft'+1)
+                              else (rAddr, amtLeft')
+        brReadAddr = rAddr'
+        -- return our new state and outputs
+        otpAck = Maybe.isJust maybePush
+        otpDat = if amtLeft_ < maxBound then Data brRead_ else NoData
```
martijnbastiaan, line 690:
```suggestion
        otpDat = if amtLeft_ < maxBound then Data brRead_ else NoData
```

Try to avoid the `variable_` pattern, as it's easy to miss when reading code.

src/Protocols/DfConv.hs
```diff
@@ -0,0 +1,1301 @@
...
+      when (burstLenLeft == 0)
+        $ put (succResizing $ fromMaybe 0 (fromKeepType $ _arlen msg), _arid msg)
+      P.pure (S2M_ReadAddress{ _arready = burstLenLeft == 0 && dfAckIn }
+             , Just (axi4ReadAddrMsgToReadAddrInfo msg))
+
+    succResizing :: (KnownNat n) => Index n -> Index (n+1)
+    succResizing n = (resize n) + 1
+
+    sendData dfDatIn dataAck = do
+      (burstLenLeft,readId) <- get
```
martijnbastiaan, line 452:
```suggestion
      (burstLenLeft, readId) <- get
```

src/Protocols/DfConv.hs
```diff
@@ -0,0 +1,1301 @@
...
+      (dataVal, dfAckOut) <- sendData dfDatIn dataAck
+      P.pure ((addrAck,dataVal),dfDatOut,dfAckOut)
+
+    processAddr M2S_NoReadAddress _
+      = P.pure (S2M_ReadAddress { _arready = False }, Nothing)
+    processAddr msg dfAckIn = do
+      (burstLenLeft,_) <- get
+      when (burstLenLeft == 0)
+        $ put (succResizing $ fromMaybe 0 (fromKeepType $ _arlen msg), _arid msg)
+      P.pure (S2M_ReadAddress{ _arready = burstLenLeft == 0 && dfAckIn }
```
martijnbastiaan, line 445:
```suggestion
      P.pure ( S2M_ReadAddress{ _arready = burstLenLeft == 0 && dfAckIn }
```

martijnbastiaan (approved):
Some small changes left, but in general LGTM. Great work!

PR #44, opened 2022-07-15
AXI4 Refactor

samalws:
This PR refactors AXI4 types so that they each only need one config type variable. This variable is a struct containing all of the `Bool` and `Nat` values that used to be passed around individually. It also adds a way to grab optional fields polymorphically as a `Maybe` value (`Nothing` if the field is turned off, `Just value` if the field is turned on).


martijnbastiaan (approved):
Much better, thanks!

PR #48, opened 2022-08-25
GHC 9.0 support

cuddlefishie:
This includes the updated `circuit-notation` plugin and some changes regarding deep subsumption.


martijnbastiaan (approved):
Thanks!

PR #50, opened 2022-08-29
check for clash-prelude version instead of GHC version

cuddlefishie:
Two instance declarations were added to `clash-protocols` in 1.7.0.
Previously the conditional inclusion was errornously based on the GHC version.


DigitalBrains1:
The feature will be incorporated into 1.6.4: [#2225](https://github.com/clash-lang/clash-compiler/pull/2225). So I think you can do:
```
#if !MIN_VERSION_clash_prelude(1, 6, 4)
```

martijnbastiaan (approved):

PR #51, opened 2022-08-30
change `wishbonePropWithModel` to use `PropertyT` instead of `Property`

cuddlefishie:


martijnbastiaan (approved):
Sure, any specific use cases?

cuddlefishie:
This came up when testing a circuit that's generic over a type level nat, the generation of the random type level nat happens inside of a `Property`/`PropertyT m ()` already. When this function is a `Property` that means it can only be a standalone property and not composed into more complex properties

PR #53, opened 2023-05-02
Change fifo implementation

lmbollen:
We now have a fifo implementation that exposes metadata called `fifoWithMeta`. The previous implementation of `fifo` has been replaced with a new one. The only difference is that the new implementation does not implement bypassing when the fifo is empty.


src/Protocols/Df.hs
```diff
@@ -635,64 +635,84 @@ registerBwd
...
+
+-- | State record for the FIFO circuit.
+data FifoState depth = FifoState
+  { readPointer   :: C.Index depth
+  , dataCount     :: C.Index (depth C.+ 1)
+  } deriving (Generic, C.NFDataX)
+
+-- | A generic First-In-First-Out (FIFO) circuit with a specified depth that exposes meta information such as a ready signal and data count.
+-- At least one cycle latency.
+-- When the reset is high or the enable is low, the output will be `Nothing` and the ready signal will be `False`.
```
martijnbastiaan, line 662: Please keep to 80 chars soft / 90 chars hard. Feel free to add

```json
  "editor.rulers": [
    80,
    90
  ],
```

to `.vscode/settings.json`.

src/Protocols/Df.hs
```diff
@@ -635,64 +635,84 @@ registerBwd
...
+fifoWithMeta ::
+  forall dom a depth .
+  (C.HiddenClockResetEnable dom,  1 C.<= depth, C.NFDataX a) =>
+  -- | The depth of the FIFO, should be at least 1.
+  C.SNat depth ->
+  -- | Consumes `Df dom a`, produces `Df dom a` along with ready signal and data count.
+    Circuit (Df dom a) (Df dom a, CSignal dom C.Bool, CSignal dom (C.Index (depth C.+ 1)))
```
martijnbastiaan, line 669:
```suggestion
  Circuit (Df dom a) (Df dom a, CSignal dom FifoMeta)
```

I don't think it's that common to export a "ready" signal. Could you export `empty`/`full` instead?

martijnbastiaan (changes requested):

martijnbastiaan:
> The previous implementation of fifo has been replaced with a new one. The only difference is that the new implementation does not implement bypassing when the fifo is empty.

Couldn't the old implementation be adapted to export empty/full signals? As it stands users of `fifo` will experience a regression.

PR #61, opened 2024-02-14
Bump `circuit-notation` version

martijnbastiaan:
TODO:

- [x] Compile with `-Werror` on CI (also see https://github.com/clash-lang/clash-protocols/pull/62)
- [x] Documentation on `Tagged` stuff
- [x] ~~Fix warnings produced by `circuit-notation`~~ No warnings on currently supported GHC versions
- [x] Module documentation
- [x] Sensible tuple sizes for instances
- [x] Fix 2-tuple documentation issue

### DIFF
--- a/src/Protocols/Plugin.hs
+++ b/src/Protocols/Plugin.hs
@@ -2,19 +2,28 @@
 A GHC source plugin providing a DSL for writing Circuit components. Credits to
 @circuit-notation@ at <https://github.com/cchalmers/circuit-notation>.
 -}
-{-# LANGUAGE OverloadedStrings #-}
 {-# LANGUAGE CPP #-}
+{-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE PatternSynonyms #-}
 
-module Protocols.Plugin where
+module Protocols.Plugin
+  ( plugin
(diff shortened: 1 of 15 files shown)


src/Protocols/Plugin.hs
```diff
@@ -2,10 +2,26 @@
...
 {-# LANGUAGE CPP #-}
+{-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE PatternSynonyms #-}
+
+module Protocols.Plugin
+  ( plugin
+  , circuit
+  , (-<)
+
+  -- XXX: Haddock will try to link references even for non-exported identifiers. We
```
kleinreact, line 14: I usually prefer [this](https://stackoverflow.com/questions/16887023/is-it-possible-to-hide-specific-functions-from-appearing-in-the-documentation-us) solution for this case, which is clean and transparent.

kleinreact (approved):
LGTM. Only some minor comment in case you care about keeping your internals internal.

src/Protocols/Plugin.hs
martijnbastiaan: Done!

PR #63, opened 2024-02-15
Support GHC 9.2, 9.4, and 9.6

martijnbastiaan:
Blocked on:

- [x] https://github.com/cchalmers/circuit-notation/pull/21


kleinreact (approved):
LGTM. 

The only thing I recommend is also adding an issue for your TODO marks in the source code, so it doesn't get lost in the woods.

martijnbastiaan:
Yeah, that's a good idea!

PR #66, opened 2024-02-28
Adding custom `Eq` instance for `Axi4StreamM2S`

lmbollen:
The derived `Eq` instance always checks the contents of each `_tdata` entry , even if the corresponding `_tkeep` entry is `False`.

According to the spec, bytes with a low keep signal, do not carry any information.
The current derived implementation leads to undesired failing comparisions:
```
              ┃   │ ━━━ Failed (- lhs) (+ rhs) ━━━
              ┃   │   Just
              ┃   │     Axi4StreamM2S {
              ┃   │         _tdata =
              ┃   │ -         0 :> 0 :> 0 :> 1 :> Nil
              ┃   │ +         0 :> 0 :> 0 :> 0 :> Nil
              ┃   │       , _tkeep =
              ┃   │           False :> False :> False :> False :> Nil
              ┃   │       , _tstrb =
              ┃   │           False :> False :> False :> False :> Nil
              ┃   │       , _tlast =
              ┃   │           True
              ┃   │       , _tid =
              ┃   │           0
              ┃   │       , _tdest =
              ┃   │           0
              ┃   │       , _tuser =
              ┃   │           ()
              ┃   │       }
```

I imagine this is also the case for some other protocols.

### DIFF
--- a/src/Protocols/Axi4/Stream.hs
+++ b/src/Protocols/Axi4/Stream.hs
@@ -83,10 +83,21 @@ deriving instance
   , C.NFDataX userType
   ) => C.NFDataX (Axi4StreamM2S conf userType)
 
-deriving instance
-  ( KnownAxi4StreamConfig conf
-  , Eq userType
-  ) => Eq (Axi4StreamM2S conf userType)
+instance ( KnownAxi4StreamConfig conf, Eq userType) => Eq (Axi4StreamM2S conf userType) where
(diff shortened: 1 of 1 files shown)


martijnbastiaan:
If you want to write a Hedgehog test for it: `a == b |-> hash(a) == hash(b)`.

lmbollen:
^ After some more thought, I'd like to change the implementation of these protocols to use ADTs. Let's discuss IRL and maybe open a discussion.

PR #67, opened 2024-03-08
Remove `CSignal` constructor

martijnbastiaan:
This is a leftover from when `Fwd` and `Bwd` were injective type families, something that didn't work out in the end. We do still need `CSignal` as a type, to work around:

https://github.com/clash-lang/clash-compiler/issues/760


src/Protocols/Internal.hs
```diff
@@ -619,10 +620,10 @@ instance (C.NFDataX a, C.ShowX a, Show a, C.KnownDomain dom) => Drivable (CSigna
   driveC _conf [] = error "CSignal.driveC: Can't drive with empty list"
   driveC SimulationConfig{resetCycles} fwd0@(f:_) =
     let fwd1 = C.fromList_lazy (replicate resetCycles f <> fwd0 <> repeat f) in
-    Circuit ( \_ -> ((), CSignal fwd1) )
+    Circuit ( \_ -> ((), fwd1) )
 
   sampleC SimulationConfig{resetCycles, ignoreReset} (Circuit f) =
-    let sampled = CE.sample_lazy ((\(CSignal s) -> s) (snd (f ((), def)))) in
+    let sampled = CE.sample_lazy (id (snd (f ((), def)))) in
```
cuddlefishie, line 626: Is this `id` needed? From what I can see it could be simplified to this, no?
```suggestion
    let sampled = CE.sample_lazy (snd (f ((), def))) in
```

martijnbastiaan:
(I'll squash the commits later.)

src/Protocols/Internal.hs
```diff
@@ -199,8 +203,8 @@ instance C.KnownNat n => Protocol (C.Vec n a) where
 
 -- XXX: Type families with Signals on LHS are currently broken on Clash:
 instance Protocol (CSignal dom a) where
-  type Fwd (CSignal dom a) = CSignal dom a
-  type Bwd (CSignal dom a) = CSignal dom ()
+  type Fwd (CSignal dom a) = Signal dom a
+  type Bwd (CSignal dom a) = ()
```
christiaanb, line 207: Why this change? why not:
```haskell
type Bwd (CSignal dom a) = Signal dom ()
```
Now `Bwd ()`  and `Bwd (CSignal dom a)` are the same, and there's even less hope to make `Bwd` injective again.

christiaanb (approved):

src/Protocols/Internal.hs
martijnbastiaan: I've given up the hope.. If we make an instance for `DSignal` (`CDSignal`? :woozy_face:) it will have the same `Bwd`.

martijnbastiaan: IMO its existence would be motivated similar to `CSignal`: sometimes you just want to deal with a non-backpressure thing in `Circuit` context.

In any case, I just remember the original motivation for dropping injectivity: all the protocols having a simple "acknowledge" on their `Bwd` would have to make newtypes, which turned out to be really annoying.

DigitalBrains1: In general, you can mix backpressure and latency just fine; it's often a requirement for efficient pipelining without buffers. You'd have the contract that when `ready` deasserts, you can still transfer _n_ more samples for a certain fixed _n_, after which the transfer stalls. For instance the Avalon Streaming interface supports this notion.

I'm not acquainted enough with `clash-protocols` to really say how this relates here; I just noticed a dichotomy being constructed between latencies and backpressure that I think isn't there.

PR #70, opened 2024-03-14
Df: `registerBwd` now places only a single register on the data path

rowanG077:
Currently `registerBwd` fully registers both the forward and backward paths. Which is somewhat misleading given the name. This PR makes it so it only breaks up the path on the backward direction. Which can be done with one register less on the forward path.

### DIFF
--- a/src/Protocols/Df.hs
+++ b/src/Protocols/Df.hs
@@ -59,6 +59,10 @@ module Protocols.Df
   , forceResetSanity
   , dataToMaybe
   , maybeToData
+  , hasData
+  , noData
+  , fromData
+  , toData
   ) where
 
 -- base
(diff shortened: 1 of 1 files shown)


rowanG077:
Resource usage for `Data (BitVector 64)` on an ECP5 using yosys

Current `registerBwd`:

```
Info:             TRELLIS_FF:   130/24288     0%
Info:           TRELLIS_COMB:    71/24288     0%
```

This PR `registerBwd`:

```
Info:             TRELLIS_FF:    65/24288     0%
Info:           TRELLIS_COMB:    67/24288     0%
```

This PR `registerBwd |> registerFwd`/ `registerFwd |> registerBwd`:

```
Info:             TRELLIS_FF:   130/24288     0%
Info:           TRELLIS_COMB:    71/24288     0%
```

martijnbastiaan (approved):
Very nice!

PR #73, opened 2024-04-03
Introduce Protocols.Df.coerce

bgamari:
This is an obvious missing combinator which I have found useful.


martijnbastiaan:
Thanks!

PR #74, opened 2024-04-10
Idle bus typeclasses

lmbollen:
Adds typeclasses that enable us to create Circuits that do nothing, can mainly be used to tie off unused inputs.
Useful for example in this case: https://github.com/bittide/bittide-hardware/pull/424/files#diff-bb29a93c87cbea176e39950d009f573873b4eaf327687b90d22a574be5619a05R104-R112

### DIFF
--- a/.ci/build_docs.sh
+++ b/.ci/build_docs.sh
@@ -20,29 +20,36 @@ cabal haddock \
   |& tee haddock_log
 
 set +e
-if grep -q "Missing documentation" haddock_log; then
+
+suppressed_warnings=(
+  'Consider exporting it together with its parent(s) for code clarity.'
+  )
+
+grep -v -e "${suppressed_warnings[@]}" haddock_log |& tee haddock_filtered
+
+if grep -q "Missing documentation" haddock_filtered; then
   echo -e "\e[1m\e[31mMissing documentation! Scroll up for full log.\e[0m"
-  grep --color=always -n -C 5 "Missing documentation" haddock_log
+  grep --color=always -n -C 5 "Missing documentation" haddock_filtered
   exit 1
 fi
 
-if grep -q "If you qualify the identifier, haddock can try to link it anyway" haddock_log; then
+if grep -q "If you qualify the identifier, haddock can try to link it anyway" haddock_filtered; then
   echo -e "\e[1m\e[31mIdentifier out of scope! Scroll up for full log.\e[0m"
-  grep --color=always -n -C 5 "If you qualify the identifier, haddock can try to link it anyway" haddock_log
+  grep --color=always -n -C 5 "If you qualify the identifier, haddock can try to link it anyway" haddock_filtered
   exit 1
 fi
 
-if grep -q "could not find link destinations for" haddock_log; then
+if grep -q "could not find link destinations for" haddock_filtered; then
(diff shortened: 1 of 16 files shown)


rowanG077:
Is there a reason all of the TH was handcrafted instead of at least partially relying on quasiquotation?

DigitalBrains1:
I always try to maximally use quotations because the code is much more readable that way. I can have a go at this code to see where I can get.

lmbollen:
@rowanG077 I'm not very experienced at writing TH, so no. @DigitalBrains1 Feel free to suggest improvements / changes :)

DigitalBrains1:
Sooo I got a version that uses TH quotations in all the places I could get them. I'm a bit disappointed; since you can't write
```haskell
f :: DecsQ
f = [d|
  g = ...
   where
    $h
  |]
 where
  h :: DecsQ
  h = ...
```
nor the equivalent variant with `g = let $h in ...`, since you are not allowed to splice a `DecsQ` like `h` there, I had to resort to building a TH expression by hand there.

My variant is at [peter/idle-bus-typeclasses](https://github.com/clash-lang/clash-protocols/tree/peter/idle-bus-typeclasses).

I wanted to achieve the following goals:
1. Use a quotation to write the instance
2. Export the function to generate the TH splice for tuples to the user

Goal (1) requires me to have the class definition in scope where I write the TH quotation. For (2) I picked the module `Protocols.Internal.TH` as a logical candidate, but unfortunately this meant I had to reorganise to avoid import loops. Also note that `.hs-boot` does not combine with Template Haskell; reorganisation is truly the only way.

So I moved the classes `Circuit`, `Protocol` and the new `IdleCircuit` to a new module `Protocols.Internal.Classes` that is _**not**_ exposed. This to prevent users accidentally importing just the class but not the instances. Instead, the API remains identical: `Circuit` and `Protocol` and their instances are available in `Protocols.Internal` and `Protocols`. So this reorganisation does not change the API, all the changes are thoroughly hidden.

This is a suggestion. We could also use a bit fewer quotations and keep everything where it was. @martijnbastiaan, I'd like to hear your opinion.

And, by the way, with the reorganisation, we could also rewrite the rather trivial hand-written TH for `protocolTupleInstances` using a bit more quotations. I could add that here if people like it.

I also changed the API for `protocolTupleInstances` and `idleCircuitTupleInstances`. Suppose a user uses the `clash-protocols` package with default tuple sizes (up to 12-tuples) but needs up to 16-tuples. Rather than change the flags for the package to generate all tuple instances, they could also use `protocolTupleInstances 13 16` to generate 13- to 16-tuples with the new API. Before, the exported `protocolTupleInstances` wasn't useful for users, because if they invoked it as `protocolTupleInstances 16`, they would get duplicate instances for 3- to 12-tuples, so they practically can't use the function. I also exported all functions in `Protocols.Internal.TH`; it is all clearly marked internal, just give these toys to the user if they want to have them.

Finally, I squashed most commits after discussion with @lmbollen. I don't see a good reason to have
* Add `IdleCircuit` typeclass
* Add `IdleCircuit (CSignal dom a)`
* Add tuple instances of `IdleCircuit`
* Add `IdleCircuit (Vec n p)`

All these can just be squashed to _Add `IdleCircuit` typeclass_.

DigitalBrains1:
> What do you think of calling the module `Idle` instead?

I have no opinion :-)

PR #75, opened 2024-05-16
Add `forceResetSanity` for all protocols

lmbollen:
`forceResetSanity` ensures combinatorially that a component does not produce data or acknowledge transactions while in Reset.

Prior to this PR, this functionality was only implemented for `Df`.
This pull requests adds a generic version based on the `IdleCircuit` typeclass to be use internally.


martijnbastiaan (approved):
Thank you. Also I'm open to PRs accepting auto formatters :eyes:

PR #76, opened 2024-05-16
Move Documentation check to separate job

lmbollen:
Currently all cabal tests can fail due to missing documentation, this is undesirable since we can not distinguish ghc / clash version related problems from missing documentation failures.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -111,10 +111,6 @@ jobs:
         run: |
           .ci/test_cabal.sh
 
-      - name: Documentation
-        run: |
-          .ci/build_docs.sh
-
   linting:
     name: Source code linting
     runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -125,3 +121,14 @@ jobs:
...
+  documentation:
+    name: Documentation clash_version 9.6.4
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Build
+        run: |
+          .ci/build_docs.sh -clash_version 9.6.4
```
martijnbastiaan, line 134:
```suggestion
          .ci/build_docs.sh --clash_version 9.6.4
```

https://serverfault.com/a/387938

martijnbastiaan (changes requested):

.github/workflows/ci.yml
```diff
@@ -125,3 +121,14 @@ jobs:
       - name: Whitespace
         run: |
           .ci/test_whitespace.sh
+
+  documentation:
+    name: Documentation clash_version 9.6.4
```
martijnbastiaan, line 126:
```suggestion
    name: Documentation
```

9.6.4 => GHC version. But also, it's fine to not mention it in the CI job name anyway.

martijnbastiaan (changes requested):

.github/workflows/ci.yml
martijnbastiaan: Also also, why are we setting Clash's version exactly?

lmbollen: Good question, I'm removing it entirely

PR #79, opened 2024-05-28
Add fourmolu as a formatter

rowanG077:
See the config for the things we can change.

A hurdle remains due to usage of CPP. Fourmolu has limited support for it which is not unsurprising which currently leads to some files not being able to be formatted. The files which can't be formatted are:

```
src/Protocols/Cpp.hs:1:1
  The GHC parser (in Haddock mode) failed:
  {ErrorWithoutFlag
   lexical error in pragma at character 'e'}
src/Protocols/Df.hs:254:1
  The GHC parser (in Haddock mode) failed:
  [GHC-58481] parse error (possibly incorrect indentation or mismatched brackets)
src/Protocols/DfConv.hs:188:1
  The GHC parser (in Haddock mode) failed:
  [GHC-58481] parse error (possibly incorrect indentation or mismatched brackets)
src/Protocols/Hedgehog.hs:57:1
  The GHC parser (in Haddock mode) failed:
  [GHC-58481] parse error (possibly incorrect indentation or mismatched brackets)
src/Protocols/Internal.hs:702:1
  The GHC parser (in Haddock mode) failed:
  [GHC-58481] parse error (possibly incorrect indentation or mismatched brackets)
```

With the exception of `src/Protocols/Cpp.hs`, these are all papering over the `unsafeFromActiveLow` and friends rename fallout. We could do several things, among others:

- Drop support for older clash versions
- Introduce a `Clash.Compat` module which papers over the difference.

Also please don't see the formatting configuration as final. It's simply what I came up with when going through all options.


rowanG077:
Oh and I also added nix based on the clash starter template.

martijnbastiaan:
I'm fine with dropping older Clash's. In fact, I wanted to drop pre GHC 9.4 too given that:

* It's the most recent GHC that is supported by Clash and has sane record handling
* Clash protocols is record heavy

martijnbastiaan:
Also damn you're quick if this was in response to my comment on the other PR.

rowanG077:
Yeah it was in response to that. But all of the work was already done for the clash ethernet core. This PR was just copy pasting things around.

fourmolu.yaml
```diff
@@ -0,0 +1,18 @@
+indentation: 2
+column-limit: 90
```
martijnbastiaan, line 2: I don't want to fight defaults too much, so I'd like to use this config:

```yaml
indentation: 2
column-limit: 90
```

Motivation for changes:

 * indentation: honestly I think people will change the locks on the office if I don't
 * column-limit: I'm a bit conflicted on this one. I'd like us to try to hit 80/90. Though sometimes it really doesn't make sense to stay within. We'll have to see how this plays out.

martijnbastiaan (changes requested):
Some questions:

 * Could you add this to CI?
 * Do you use this in combination with commit hooks or something?

Besides that I wish it could just ban star imports :eyes:

rowanG077:
I'm fine with anything really. I care more about it being consistent and also to minimize bike shedding during review. Could you tag anyone else at qbaylogic who will probably have an opinion on this? Perhaps @christiaanb @DigitalBrains1? I'd like to hash out the config and then make a PR with that some config to clash-compiler. I want to avoid having to do fight the good fight twice so to speak.

martijnbastiaan:
I'm in the same boat. I'd like:

* Consistency
* Not to think about style details too much
* Not to waste time on bike shedding
* To minimize onboarding time for starters

Which is also why I'm proposing to change as few configuration options as reasonable (even though I think some non-defaults are prettier). I'd like the same for clash-compiler, though I feel @christiaanb holds the keys to that.

DigitalBrains1:
The problem with enforcing it, like in the pre-commit hook, is that sometimes the result that an algorithm comes up with might be very suboptimal. I haven't used fourmolu yet (it was on my TODO list), so I don't know if it is the case in practice, but I've fought with formatters before where it produced something that just looked bad for no good reason... I mean, I could understand why the algorithm came up with it, it's just not what a practised human would produce.

rowanG077:
I think the point is that it's better to accept sub-optimal output rather then have inconsistent styling. As in favor automated consistency over perfect handcrafted human formatting.  

In extreme cases there are also annotations that disable automatic formatting between two special comments.

nix/sources.json
```diff
@@ -0,0 +1,102 @@
...
+        "description": "Haskell to VHDL/Verilog/SystemVerilog compiler",
+        "homepage": "https://clash-lang.org/",
+        "owner": "clash-lang",
+        "repo": "clash-compiler",
+        "rev": "3f5dc67c0e526e43a4dd88eb3902e39ed512c166",
+        "sha256": "022rwif8xkaabw0j3arhyj0hswmh3vq2nx1bbz8rbkp05jm4psgg",
+        "type": "tarball",
+        "url": "https://github.com/clash-lang/clash-compiler/archive/3f5dc67c0e526e43a4dd88eb3902e39ed512c166.tar.gz",
+        "url_template": "https://github.com/<owner>/<repo>/archive/<rev>.tar.gz",
+        "version": "1.8.1"
```
DigitalBrains1, line 13: Do we want this repository to track Clash stable by default? I don't know, I'm asking :-).

DigitalBrains1:
> No formatter will reach a human spending hours to format the code.

Hours? o_o

I don't want to do a whole discussion. I'm personally against enforcing this on `clash-compiler`; I don't really care about `clash-protocols` (might regret that statement later :-D). That said, I won't stop it if I'm in the minority, I'll just swallow it and follow.

rowanG077:
Well of course not for a single change. But I don't doubt hours and hours were spend on the while clash project on formatting. At least I can say I have spend a lot of time on it.

DigitalBrains1:
> At least I can say I have spend a lot of time on it.

I am absolutely not against people using fourmolu to format their code! Please by all means go ahead! My only objection regards forcing people to use it, either through policy or through CI or whatnot.

martijnbastiaan:
I've also lost many pointless hours over style back-and-forths. But mostly, I think our approach to styling is just not working. I think of all the people that work on the Clash compiler and related repos, you're (@DigitalBrains1) are the basically the only one that doesn't fall into one of these categories:

1. People who are too new to Haskell/Clash to properly apply the style guide
1. People who think styling isn't important
1. People who don't like our style guide, and stubbornly submit non-compliant code even after being told multiple times that it's not compliant. (Not excluding the maintainers themselves.)
1. People for whom the style guide is too difficult to follow properly

I've tried to be consistent and vigilant for this repo, but if you dig around it now it should be obvious that it's not consistent at all. So while I appreciate this concern and even agree with it:

> I've fought with formatters before where it produced something that just looked bad for no good reason... I mean, I could understand why the algorithm came up with it, it's just not what a practised human would produce.

I feel that in practice the combination of time pressure, different preferences, and human behavior just leads to very suboptimal styling when left unenforced. Not to mention that new insights and Haskell features can even lead to different preferred styling (e.g., leading vs trailing arrows), leaving even a single repository in a split on what styling to pick.

All in all, I'd like to treat styling the same as we treat warnings:

- Enforced "warning free" on CI
- Selectively disabled where it doesn't make sense

nix/sources.json
rowanG077: Copied from the clash starter. I don't spend any time thinking about it.

DigitalBrains1:
You bring forth solid arguments.

Let me emphasise a wrinkle though: it requires a "flag day" commit that brings all of the existing code into conformance with the fourmolu-mandated style. This is what happens in this PR as well. On `clash-compiler`, you'll also have to convince the maintainers this is warranted to achieve the goal of CI-enforced style. Frankly, personally I'm not against that! It is a source of irritation to me that I have to adapt my style to the specific file I'm editing because our style guide changed a few times and different files follow different eras of style. That is hard on muscle memory! This is definitely a case where I also spend actual time on style. But others might object that this makes `git blame` harder to use (pinging @christiaanb; I think this came up before in discussion).

rowanG077:
Regarding the blame issue. You can add a `.git-blame-ignore-revs` file which git picks up and understands to ignore changes from a specific commit.

DigitalBrains1:
Haha, yeah, I was just testing it with GitHub because their blame interface is nice. [It works](https://docs.github.com/en/repositories/working-with-files/using-files/viewing-a-file#ignore-commits-in-the-blame-view), I tested it in a private repo just now.

martijnbastiaan:
I was worried that it would just silently hide it, but that's not the case. Well done GitHub :+1:

christiaanb:
I'm good with 2-space-indent fourmulu defaults for `clash-compiler` as long as we use `.git-blame-ignore-revs` now that Github supports it.

rowanG077:
Fourmolu now runs on CI. I removed support for Clash < 1.8.1 because of the reset changes and also removed support for GHC < 9. I also added a `format.sh` script which you can use to locally format the code.

format.sh
```diff
@@ -0,0 +1,38 @@
+#!/usr/bin/env bash
+set -euo pipefail
```
martijnbastiaan, line 2: Could you add a `docopt` like description?

martijnbastiaan (approved):

lmbollen (approved):

PR #83, opened 2024-06-28
Add `ReqResp` Protocol

lmbollen:
A simple protocol for request-response communication. The forward channel channel has type `Signal dom (Maybe req)` and is used to send requests, the backward channel has type `Signal dom (Maybe resp)` and is used to send responses.

The protocol must obey the following rules:
 * When the forward channel is `Just a`,  it must not change until the transaction is completed.
 * The forward channel can not depend on the backward channel.
 * When the forward channel is `Nothing`, the backward channel may be undefined.
Currently only the basic protocol definition is there, I'm looking for feedback and extra desired functionality.

### DIFF
--- a/src/Protocols/ReqResp.hs
+++ b/src/Protocols/ReqResp.hs
@@ -1,0 +1,53 @@
+{- |
+Simplest possible protocol for request-response communication.
+
+The forward channel channel has type @Signal dom (Maybe req)@ and is used to send requests.
+The backward channel has type @Signal dom (Maybe resp)@ and is used to send responses.
+
+The protocol must obey the following rules:
+* When the forward channel is @Just a@, it must not change until the transaction is completed.
+* The forward channel can not depend on the backward channel.
+* When the forward channel is @Nothing@, the backward channel may be undefined.
+
+This protocol can not be pipelined, for pipelined request-response communication see `Protocols.BiDf`.
+-}
+module Protocols.ReqResp where
+
+import qualified Clash.Prelude as C
+import Data.Kind (Type)
+import Protocols
+import Protocols.Internal.Classes
+import Prelude as P
+
+{- |
+Simplest possible protocol for request-response communication.
+
+The forward channel channel has type @Signal dom (Maybe req)@ and is used to send requests.
+The backward channel has type @Signal dom (Maybe resp)@ and is used to send responses.
+
+The protocol must obey the following rules:
+* When the forward channel is @Just a@, it must not change until the transaction is completed.
+* The forward channel can not depend on the backward channel.
+* When the forward channel is @Nothing@, the backward channel may be undefined.
+-}
+data ReqResp (dom :: C.Domain) (req :: Type) (resp :: Type)
+
+instance Protocol (ReqResp dom req resp) where
+  -- \| Forward channel for ReqResp protocol:
+  type Fwd (ReqResp dom req resp) = C.Signal dom (Maybe req)
+
+  -- \| Backward channel for ReqResp protocol:
+  type Bwd (ReqResp dom req resp) = C.Signal dom (Maybe resp)
+
+instance IdleCircuit (ReqResp dom req resp) where
+  idleFwd _ = pure Nothing
+  idleBwd _ = pure Nothing
+
+{- | Force a @Nothing@ on the backward channel and @Nothing@ on the forward
+channel if reset is asserted.
+-}
+forceResetSanity ::
+  forall dom req resp.
+  (C.HiddenReset dom) =>
+  Circuit (ReqResp dom req resp) (ReqResp dom req resp)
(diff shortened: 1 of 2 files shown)


martijnbastiaan (commented):
In general I don't see an issue with this protocol, but I'm do wondering if there is a way where we can formalize a little bit what kind of protocols there are / can be. Right now we're kinda slapping on protocol after protocol.

src/Protocols/ReqResp.hs
```diff
@@ -0,0 +1,46 @@
+{- |
+Simple protocol for request-response communication.
+The forward channel channel has type @Signal dom (Maybe req)@ and is used to send requests.
+The backward channel has type @Signal dom (Maybe resp)@ and is used to send responses.
+The protocol must obey the following rules:
+* When the forward channel is @Just a@, it must not change until the transaction is completed.
```
Losiek, line 6: This implies that this protocol can not be pipelined. We can relax this constraint and do two versions of ReqResp (pipelined and not pipelined). Or we can use DSignal to enforce different pipeline depths on the type level.

rowanG077: Why couldn't it be pipelined? I think you can write a skid buffer that handles this. Essentially store forward in a register. Send contents of the register to the sink. When the sink responds either send it directly to the source and clear the register or put the response in a register as well.

christiaanb: Perhaps we should then just have
```haskell
data ReqResp (dom :: C.Domain) (req :: Type) (resp :: Type) = (Df dom req, Reverse (Df dom resp))
```
a kind of Axi4LiteLite

DigitalBrains1: Enforcing pipeline depth on the type level seems to be more useful than just a blanket declaration "you can send stuff, which we call a request, and you can receive stuff, which we call a response", which seems to be what Christiaans `ReqRespPipelined` is.

Avalon Stream does pre-negotiated pipelining. You can configure that when you drop _ready_, you can still receive data in the next _n_ cycles, so the pipeline of the sender can empty. Constructions like that. Maybe it can serve as inspiration.

Alternatively, we can tackle pipelining in a different PR.

t-wallet:
Looks good. This is the generic version of the `ArpLookup` protocol (to be upstreamed to `clash-cores`), so we can nicely specify that in terms of `ReqResp`. We did not need more functionality than just the protocol definition there, so I don't have anything to add for now.

src/Protocols/ReqResp.hs
lmbollen: I don't think separating the channels is a good idea as it directly contradicts the purpose of `circuit-notation` (Combining forward and backward channels in a single binder).

How about separating request acknowledgement and response timing?:
```hs
data ReqResp (dom :: C.Domain) (req :: Type) (resp :: Type)

instance Protocol (ReqResp dom req resp) where
  -- | Forward channel for ReqResp protocol
  type Fwd (ReqResp dom req resp) = C.Signal dom (Maybe req)

  -- | Backward channel for ReqResp protocol
  type Bwd (ReqResp dom req resp) = C.Signal dom (Bool, Maybe resp)
  ```
Here the `Bool` would indicate a request acknowledgement and `Maybe resp` could be `Just resp` later.
Combinatorial circuits could then return `(isJust resp, resp)`.

DigitalBrains1: How would you still be able to tell which response corresponds to which request?

Personally, I think a "simple" protocol that doesn't allow pipelining at the protocol level¹ and a separate more complex protocol that tackles pipelining in a neat and efficient way is a perfectly fine solution. This PR can be the simple variant.

¹ As Rowan indicates above, you can still accomodate pipelining with constructions like skid buffers, but the protocol would be oblivious to this and it would in general be less efficient than pipelining being part of the protocol.

bgamari:
It turns out that I have done something very similar to this in !109.

lmbollen:
> It turns out that I have done something very similar to this in #109.

Great minds think alike it seems ;)

Your pull request already has tests so I'm inclined to continue with that one.
One thing I'm missing is the distinction between `BiDf` masters and `BiDf` slaves.
Shall I close this PR in favour of yours?

DigitalBrains1:
> One thing I'm missing is the distinction between BiDf masters and BiDf slaves.

So I'm not really a `clash-protocols` guy, still need to build some experience, but I don't really follow what you mean. `BiDf` is just a bidrectional channel with backpressure, right? Participant A could decide to send 42 data units to participant B and B could decide to stay silent this run or just randomly start spewing 1337 data units to participant A. I don't think there's any definition beyond there being two participants. So what in your eyes is a manager and what is a subordinate? (Or some other term if you prefer, but please no masters or slaves).

PR #85, opened 2024-07-03
Add PacketStream protocol.

rowanG077:
Todos:

- [ ] Checkout haddock and make nicer
- [ ] Copy over tests
- [ ] Make top-level PacketStream module that exports all sub-modules


src/Protocols/PacketStream/AsyncFifo.hs
```diff
@@ -0,0 +1,60 @@
...
+  -- | Enable signal in the write domain
+  Enable wDom ->
+  -- | Clock signal in the read domain
+  Clock rDom ->
+  -- | Reset signal in the read domain
+  Reset rDom ->
+  -- | Enable signal in the read domain
+  Enable rDom ->
+  Circuit (PacketStream wDom dataWidth metaType) (PacketStream rDom dataWidth metaType)
+asyncFifoC depth wClk wRst wEn rClk rRst rEn = fromSignals ckt
```
clash-lang-builder, line 49: This circuit behaves weird. I'm quite used to just put `enableGen` on my global enable lines. However, when I do that for something involving this asyncFifoC while the reset is asserted, I'll basically be writing garbage into the fifo. That's because the `PacketStream` does not have an inherent validity, and since the fifo won't be full, the ack will be asserted.

christiaanb: But I guess it's not a problem as the write pointer won't make any progress.

christiaanb: It does however depend on having a single reset for the entire domain, which is not mandatory.

src/Protocols/PacketStream/Base.hs
```diff
@@ -0,0 +1,213 @@
...
+import Data.Coerce (coerce)
+import qualified Data.Maybe as Maybe
+import Data.Proxy
+
+{- | Data sent from manager to subordinate, a simplified AXI4-Stream like interface
+with metadata that can only change on packet delineation.
+_tdest, _tuser and _tid are bundled into one big _meta field which holds metadata.
+There are no null or position bytes so _tstrb is replaced by a last indicator
+that indicates the index of the last valid byte in the _data vector.
+_tvalid is modeled via wrapping this in a `Maybe`.
```
christiaanb, line 38: A `_tvalid` is needed in case the manager is released from its reset later than the subordinate it is talking to. Multiple resets is perfectly fine from a metastability point of view when you have synchronous resets. And when you have components which you think do handshaking, then it's not too weird to have different reset deassert times to achieve timing closure.

src/Protocols/PacketStream/AsyncFifo.hs
rowanG077: What do you mean `PacketStream` does not have inherent validity?

christiaanb: There is no `_tvalid` like in AxiStream. I guess the `abort` could serve that function? Should we then mandate that abort must be asserted when the component is in reset?

rowanG077: It's wrapped in a `Maybe`, So a `Nothing` means it's invalid. `tvalid` is modelled exactly like this in our clash-protocols axi stream protocol as well.

christiaanb: Ah, I missed that.

christiaanb: Apologies for the misunderstanding.

src/Protocols/PacketStream/Base.hs
christiaanb: I thought the “wrapping” in a Maybe was optional. But I now see it is part of the Protocol instance. Apologies for the misunderstanding.

src/Protocols/PacketStream/AsyncFifo.hs
```diff
@@ -0,0 +1,60 @@
...
+ where
+  ckt (fwdIn, bwdIn) = (bwdOut, fwdOut)
+   where
+    (element, isEmpty, isFull) = asyncFIFOSynchronizer depth wClk rClk wRst rRst wEn rEn readReq fwdIn
+    notEmpty = not <$> isEmpty
+    -- If the FIFO is empty, we output Nothing. Else, we output the oldest element.
+    fwdOut = toMaybe <$> notEmpty <*> element
+    -- Assert backpressure when the FIFO is full.
+    bwdOut = PacketStreamS2M . not <$> isFull
```
christiaanb, line 58: Data will be lost when the fifo is held in reset. Perhaps `asyncFifoSynchronizer` should be changed so that it asserts `full` when the write side is in reset.

rowanG077: True. `forceResetSanity` will fix that as well.

src/Protocols/PacketStream/Base.hs
```diff
@@ -0,0 +1,213 @@
...
+
+deriving instance
+  (KnownNat dataWidth, NFDataX metaType) =>
+  NFDataX (PacketStreamM2S dataWidth metaType)
+
+instance Protocol (PacketStream dom dataWidth metaType) where
+  type
+    Fwd (PacketStream dom dataWidth metaType) =
+      Signal dom (Maybe (PacketStreamM2S dataWidth metaType))
+  type Bwd (PacketStream dom dataWidth metaType) = Signal dom PacketStreamS2M
```
leonschoorl, line 72: The flow control signals of `PacketStream` are isomorphic to those of `Df`.
And going on your explanation [here](https://github.com/clash-lang/clash-protocols/discussions/84#discussioncomment-9948153) I think the meaning of the flow control signals are basically the seems between the two protocols.
Would it make sense to reuse those `Df` data types in `PacketStream`?

```haskell
instance Protocol (PacketStream dom dataWidth metaType) where
  type Fwd (PacketStream dom dataWidth metaType) =
      Signal dom (Data (PacketStreamM2S dataWidth metaType))
  type Bwd (PacketStream dom dataWidth metaType) = Signal dom Ack
```

rowanG077: Yes that's possible. We didn't do it because `AxiStream` protocol also didn't do it. We just followed that as our baseline.

rowanG077: We will move over `Df` to use `Maybe`, see https://github.com/clash-lang/clash-protocols/issues/87

bgamari:
The motivation here is a bit unclear to me. The documentation says "Provides the PacketStream protocol, a simple streaming protocol for transferring packets of data between components" however the "data" in question appears to be only bytes. Presumably there is a specific use-case in mind here but it is not clear to me what it is; network packets, perhaps?

It seems to me like ideally protocols in `clash-protocols` should be polymorphic in the type of the data that they transfer. I wouldn't expect packets of bytes to be terribly prevalent in most design as surely you would want to parse the packet into a more structured form somewhere near the "edge" of the design.

t-wallet:
> The motivation here is a bit unclear to me. The documentation says
> 
> > Provides the PacketStream protocol, a simple streaming protocol for transferring packets of data between components
> 
> However the "data" in question appears to be only bytes. Presumably there is a specific use-case in mind here but it is not clear to me what it is; network packets, perhaps?
> 
> It seems to me like ideally protocols in `clash-protocols` should be polymorphic in the type of the data that they transfer. Afterall, I wouldn't expect packets of bytes to be terribly prevalent in the "interior" of most designs. Unless you are building actual networking hardware, I would imagine you would typically want to parse the packet into a more structured form somewhere near the "edge" of the design.

This protocol was indeed specifically created for a hardware-based MAC/IP/UDP stack (which will soon be coming to `clash-cores`). However, it is just a simpler version of AXI Stream with stricter metadata invariants. You can use it for any other purpose that needs high-speed data streaming, like video streaming. This PR also provides several circuits which are able to read these unstructured bytes into arbitrary types.

rowanG077:
It was discussed with @lmbollen whether to make the payload polymorphic. In theory there isn't anything that currently really looks at the payload so it would not be super hard to do add a type variable for the payload.

But in my experience working with packetized streams there usually is not a sane way to have an interior type. You are almost certainly only dealing with fragments, these will generally not have a clear structure at the scope of a single cycle. The structure comes from the fragments together. 

Do you have some concrete idea how you see a packet stream packet with a non-byte oriented payload?

martijnbastiaan:
>  But in my experience working with packetized streams there usually is not a sane way to have an interior type. 

This is my experience too. In particular, any protocol that has byte enables (including the interface of block RAMs) are a pain to deal with if they're not having bits/bytes/bitvector as an interior type.

Still, ADTs are _the_ killer feature of Clash IMO, so perhaps we should also offer API compatible versions without byte enables.

t-wallet:
@rowanG077
I've done everything I wanted to do. The PR is ready for a final review.

clash-protocols/src/Clash/Sized/Vector/Extra.hs
```diff
@@ -0,0 +1,59 @@
...
+ where
+  go :: forall l. SNat l -> Vec (n + m) a
+  go l@SNat =
+    let f = addSNat l d1
+     in case compareSNat f (SNat @n) of
+          SNatLE -> takeLe (addSNat l d1) xs ++ ys ++ extra
+           where
+            extra :: Vec (n - (l + 1)) a
+            extra = repeat 0
+          _ -> error "appendVec: Absurd"
```
rowanG077, line 58: `clashCompileError "..."`

clash-protocols/clash-protocols.cabal
```diff
@@ -77,6 +77,7 @@ common common-options
     -fplugin GHC.TypeLits.Extra.Solver
     -fplugin GHC.TypeLits.Normalise
     -fplugin GHC.TypeLits.KnownNat.Solver
+    -fconstraint-solver-iterations=6
```
rowanG077, line 80: I would rather move this to the offending module(s). This is a bit of a sledgehammer.

clash-protocols/src/Protocols/PacketStream/Base.hs
```diff
@@ -0,0 +1,554 @@
...
+  forall n dataWidth meta dom.
+  (HiddenClockResetEnable dom) =>
+  (KnownNat n) =>
+  (KnownNat dataWidth) =>
+  (1 <= n) =>
+  (NFDataX meta) =>
+  Circuit
+    (PacketStream dom dataWidth meta)
+    (Vec n (PacketStream dom dataWidth meta))
+fanout = DfConv.fanout Proxy Proxy
```
rowanG077, line 380: This doesn't preserve expected `PacketStream` semantics. It will distribute fragments of the same packet over multiple consumers. Something you most likely do not want to do. This should be explicit in the documentation. Persoanally I wouldn't export it all specialized for `PacketStream`

clash-protocols/src/Protocols/PacketStream/Base.hs
```diff
@@ -0,0 +1,554 @@
...
+-- | Like 'Data.Bifunctor.second', but over the metadata of a 'PacketStream'.
+secondMeta ::
+  (B.Bifunctor p) =>
+  (b -> c) ->
+  Circuit
+    (PacketStream dom dataWidth (p a b))
+    (PacketStream dom dataWidth (p a c))
+secondMeta f = secondMetaS (pure f)
+
+-- | Like 'secondMeta', but can reason over signals.
```
rowanG077, line 547: Perhaps as @DigitalBrains1 noted we should make clear that these `*S` circuit combinators are akin to `<*>`.

rowanG077:
I have not reviewed everything yet. I will take some time this week to look everything over in detail.

clash-protocols/src/Clash/Sized/Vector/Extra.hs
t-wallet: I just removed the function as it is unused. It is very specific and was only used in FCS insertion.

clash-protocols/src/Protocols/PacketStream/Routing.hs
```diff
@@ -0,0 +1,93 @@
...
+import qualified Data.Bifunctor as B
+import Data.Maybe
+
+-- | Collect mode for `packetArbiterC`
+data ArbiterMode
+  = -- | Collect in a round-robin fashion. Fair and cheaper than `Parallel`.
+    RoundRobin
+  | -- | Check components in parallel. This mode has a higher throughput, but is
+    -- biased towards the last source and also slightly more expensive.
+    Parallel
```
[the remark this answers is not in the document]
t-wallet: Just checked it, it's not quite the same. `roundrobinCollect` also has a `NoSkip` mode:
```
Collect in a roundrobin fashion. If a component does not produce data, wait until it does.
```
We could handle add this mode to the arbiter as well.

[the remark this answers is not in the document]
lmbollen: That would be nice, let's try to make it uniform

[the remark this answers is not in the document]
t-wallet: Done. Added a fat warning to the `NoSkip` mode because it can easily lead to starvation. In fact, the test can only successfully run if the same amount of packets are generated for each source.

PR #91, opened 2024-07-20
Do not verify lengths in `propWithModel`

### LINKED ISSUE #81: `propWithModel` and friends should not verify lengths
From the documentation:
```hs
-- | Test a protocol against a pure model implementation. Circuit under test will
-- be arbitrarily stalled on the left hand and right hand side and tested for
-- a number of properties:
--
--   * Whether it does not produce too little data.
--   * Whether it does not produce /more/ data than expected.
--   * Whether it responds to backpressure correctly
--   * Whether it (eventually) drives a /nack/ while in reset.
--
-- Finally, the data will be tested against the property supplied in the last
-- argument.
--
```

The first two should not be checked or be made optional. Currently your model has to exactly mimic the implementation regarding the production of transactions. This can require larger than desired efforts for no gain.

For example.
An `Axi4 Stream` 1 to n resizing component could receive transactions such as the ones below:
```hs
[ Axi4StreamM2S
    { _tdata = 0 :> Nil
    , _tkeep = True :> Nil
    , _tstrb = True :> Nil
    , _tlast = False
    , _tid = 0
    , _tdest = 0
    , _tuser = ()
    }
, Axi4StreamM2S
    { _tdata = 0 :> Nil
    , _tkeep = True :> Nil
    , _tstrb = True :> Nil
    , _tlast = False
    , _tid = 0
    , _tdest = 0
    , _tuser = ()
    }
, Axi4StreamM2S
    { _tdata = 0 :> Nil
    , _tkeep = False :> Nil
    , _tstrb = False :> Nil
    , _tlast = True
    , _tid = 0
    , _tdest = 0
    , _tuser = ()
    }
]
```
If the implementation produces:
```hs
[ Axi4StreamM2S
    { _tdata = 0 :> 0 :> Nil
    , _tkeep = True :> True :>Nil
    , _tstrb = True :> True :> Nil
    , _tlast = False
    , _tid = 0
    , _tdest = 0
    , _tuser = ()
, Axi4StreamM2S
    { _tdata = 0 :> 0 :> Nil
    , _tkeep = False :> False :>Nil
    , _tstrb = False :> False :> Nil
    , _tlast = True
    , _tid = 0
    , _tdest = 0
    , _tuser = ()
    }
``` 
And the model produces:
```hs
[ Axi4StreamM2S
    { _tdata = 0 :> 0 :> Nil
    , _tkeep = True :> True :>Nil
    , _tstrb = True :> True :> Nil
    , _tlast = True
    , _tid = 0
    , _tdest = 0
    , _tuser = ()
```
I can still verify that it correctly communicated the packet, even though the model slightly deviates. However, `propWithModel` does not allow it, requiring me to mimic exactly the way my implementation merges transactions. Sometimes you just want to verify properties of the output without supplying model, making this restriction even more annoying.

In my opinion this should not be the default behavior because it makes it significantly harder to check simple properties.

martijnbastiaan:
The input length is not always equal to the output length, nor even predictable. Furthermore, some properties might not be interested in the length in the first place.

Fixes #81

### DIFF
--- a/src/Protocols/Hedgehog.hs
+++ b/src/Protocols/Hedgehog.hs
@@ -41,8 +41,10 @@ import Protocols.Hedgehog.Internal
 -- clash-prelude
 import qualified Clash.Prelude as C
 
--- hedgehog
+-- clash-prelude-hedgehog
+import Clash.Hedgehog.Sized.Vector (genVec)
 
+-- hedgehog
 import Hedgehog ((===))
 import qualified Hedgehog as H
 import qualified Hedgehog.Gen as Gen
@@ -59,6 +61,16 @@ resetGen :: (C.KnownDomain dom) => Int -> C.Reset dom
 resetGen n =
   C.unsafeFromActiveHigh
     (C.fromList (replicate n True <> repeat False))
+
+smallInt :: H.Range Int
+smallInt = Range.linear 0 10
+
+genSmallInt :: H.Gen Int
+genSmallInt =
+  Gen.frequency
+    [ (90, Gen.integral smallInt)
+    , (10, Gen.constant (Range.lowerBound 99 smallInt))
+    ]
 
 {- | Test a protocol against a pure model implementation. Circuit under test will
 be arbitrarily stalled on the left hand and right hand side and tested for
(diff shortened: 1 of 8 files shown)


src/Protocols/Hedgehog.hs
```diff
@@ -60,6 +62,16 @@ resetGen n =
...
+smallInt :: H.Range Int
+smallInt = Range.linear 0 10
+
+genSmallInt :: H.Gen Int
+genSmallInt =
+  Gen.frequency
+    [ (90, Gen.integral smallInt)
+    , (10, Gen.constant (Range.lowerBound 99 smallInt))
+    ]
+
```
lmbollen, line 74: Consecutive stalling up to only 10 cycles is a very low default.
This should be configurable or derived from configuration? Could you make an issue and point to it here?

lmbollen (approved):
Lgtm modulo some small remarks.
Please remember to make an issue for the maximum number of consecutive stalls.

src/Protocols/Hedgehog/Internal.hs
```diff
@@ -113,58 +106,30 @@ class
...
-    go timeout n (Df.NoData : as) = do
-      -- Circuit did not output valid cycle, just continue
-      go (pred timeout) n as
-    go _ n (Df.Data _ : as) =
-      -- Circuit produced a valid cycle, reset timeout
-      go (fromMaybe maxBound eoTimeout) (pred n) as
+    go 0 _ _ =
+      H.failWith
+        Nothing
+        ("Sample limit reached after sampling " <> show eoSampleMax <> " samples")
```
martijnbastiaan, line 126: Offer potential solution

src/Protocols/Hedgehog.hs
martijnbastiaan: Added! https://github.com/clash-lang/clash-protocols/issues/98

PR #106, opened 2024-08-20
Reexport Protocols.Plugin from clash-protocols

bgamari:
This ensures that the package split introduced in e2f4da3ad5fa9218fd4506c9d433d9adb5c7bfe8 isn't visible to the user.


t-wallet (approved):
Whoops, that was a simple oversight. Thank you!

PR #109, opened 2024-09-09
Introduce BiDf protocol

bgamari:
This introduces a new protocol, `BiDf`, which captures the common pattern of a request `Df` channel paired with a `Df` for returning corresponding responses.

### DIFF
--- a/clash-protocols/src/Protocols/BiDf.hs
+++ b/clash-protocols/src/Protocols/BiDf.hs
@@ -13,6 +13,8 @@ module Protocols.BiDf (
   loopback,
     -- * Mapping
   dimap,
+  -- * Fan-in
+  fanin
 ) where
 
 import Prelude ()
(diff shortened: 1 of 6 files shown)


bgamari:
At the moment `prop_fanin` does not pass for reasons that I haven't yet understood.

christiaanb:
There seems to be some overlap with https://github.com/clash-lang/clash-protocols/pull/83 ...

lmbollen:
We're trying to be sensible about which protocols we add to this package, so now I also added `BiDf` to this overview:
https://github.com/clash-lang/clash-protocols/discussions/84

clash-protocols/src/Protocols/BiDf.hs
```diff
@@ -0,0 +1,139 @@
...
+import qualified Protocols.Df as Df
+
+-- | A 'Protocol' allowing requests to be passed downstream, with corresponding
+-- responses being passed back upstream. Responses are provided in the order that
+-- their corresponding requests were submitted.
+--
+-- *Correctness conditions*
+--
+--  - The response channel must not produce a value before the request channel
+--    has produced a value.
```
DigitalBrains1, line 34: I think it would be good to really spell out what _produced_ and _before_ mean in this sentence.

Scenario: In cycle N, the forward channel asserts `Data Request`. The recipient doesn't yet acknowledge this, but it starts to work on the response. Then in cycle N+1, it both acknowledges `Data Request` and also produces, on the backward channel, `Data Response`.

If we say the moment it is acked, the channel has produced a value (before that, the _component_ could well be said to have produced a value, but the _channel_ is really up to interpretation), we now have the situation where the response is produced at the exact same moment as the request is produced. So it wasn't _before_, but as a reader of this documentation, I am left wondering if this is the correct interpretation.

The scenario seems to be desired behaviour, so let's rephrase or give an example to really show that this is within spec.

clash-protocols/src/Protocols/BiDf.hs
```diff
@@ -0,0 +1,141 @@
...
+--
+--  - Each request must be paired with exactly one response.
+--
+--  - Responses must be issued in the order that their corresponding requests arrived.
+--
+--  - Both the request and response channels must obey usual 'Df' correctness
+--    conditions.
+--
+--  - There must not be a combinational path from the request channel to the
+--    response channel.
```
lmbollen, line 46: Am I reading correctly that this disallows stateless circuits such as:
```hs
idBiDf :: Circuit (BiDf dom a a) ()
idBiDf = Circuit go
 where
  go ((fwd, bwd), _) = ((bwd, fwd), ())
```

If so, why would that not be allowed?

PR #112, opened 2024-09-20
Add Vec and unit type instances to IdleCircuit

Linouth:
This enables attaching an idleCircuit to a Vec of protocols and to a Unit type protocol. 
This is useful when using `vecCircuits` to attach multiple slaves to an interconnect.


lmbollen (approved):
LGTM, thanks :)

PR #113, opened 2024-09-23
Fix fourmulu check

lmbollen:
Paths were broken after separating `clash-protocols` in 2 packages


rowanG077:
If we change that I would say let's also change the formatting script:

https://github.com/clash-lang/clash-protocols/blob/0832a422e77422739401896f6612620d17baa289/format.sh#L37

PR #123, opened 2024-11-08
Set version bounds for `data-default`

t-wallet:
Fixes CI failures complaining about an instance `Default` for `SimulationConfig` that could not be deduced. `data-default` version 0.8 introduced a breaking change because it does not export `Default` from `data-default-class` anymore, which is necessary for compatibility with `clash-prelude` for the moment.

### DIFF
--- a/clash-protocols-base/clash-protocols-base.cabal
+++ b/clash-protocols-base/clash-protocols-base.cabal
@@ -102,7 +102,6 @@ library
 
   build-depends:
     , circuit-notation
-    , data-default-class
     , deepseq
     , extra
     , ghc >= 8.7 && < 9.7
--- a/clash-protocols-base/clash-protocols-base.cabal
+++ b/clash-protocols-base/clash-protocols-base.cabal
@@ -102,7 +102,7 @@ library
 
   build-depends:
     , circuit-notation
-    , data-default
+    , data-default-class
     , deepseq
     , extra
     , ghc >= 8.7 && < 9.7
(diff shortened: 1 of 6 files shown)


DigitalBrains1:
For now, I'd just constrain the dependency:
```diff
--- a/clash-protocols/clash-protocols.cabal
+++ b/clash-protocols/clash-protocols.cabal
@@ -116,7 +116,7 @@ library
     , clash-protocols-base
     , circuit-notation
     , clash-prelude-hedgehog
-    , data-default
+    , data-default ^>= 0.7.1.1
     , deepseq
     , extra
     , ghc >= 8.7 && < 9.7
```
and be done with it here. And then we need to decide in `clash-compiler` how to deal with the changed relationship between `data-default` and `data-default-class`. If we want things like `clash-protocols` and `clash-cores` to potentially be compatible with both Clash 1.8 and Clash 1.10, we'll need to think about how that should work. Or if we just want compatibility with Clash 1.10, we can switch everything to `data-default >= 0.8`.

Why 0.7.1.1? Well, that's what is in several "repositories" like NixOS and the Stack used in CI, and it's also what was the current version when `clash-protocols` was created.

DigitalBrains1 (approved):
Could you squash commit with this commit msg:
```
Set version bounds for `data-default` (#123)

data-default 0.8 introduced a breaking change: it doesn't export Default
from data-default-class anymore. As clash-prelude uses
data-default-class, this leads to two incompatible versions of the
Default class. We constrain data-default for now and later decide what
all our packages should do.
```

PR #124, opened 2024-11-20
Use more quotations in TH module

DigitalBrains1:
We can use more quotations and less "bare" TH constructions.

This was intended as a change to #117, but it was already merged.


lmbollen:
I ah I see that your code too gets the unused pattern binding warning in TH.... I had them too and a a small rewrite fixed it for me, but I don't know when that triggers...

DigitalBrains1:
Yes, and the minimal fix is really silly. Instead of writing

```haskell
[d|
  Circuit $(varP ...) = ...
  |]
```

we wrap it as

```haskell
[d|
  $[p| Circuit $(varP ...) |] = ...
  |]
```

where we wrap a pattern `P` as `$[p| P |]` .

PR #110, opened 2024-09-09
Allow building with GHC 9.10

bgamari:
Requires https://github.com/cchalmers/circuit-notation/pull/26.

PR #130, opened 2025-01-27
Support GHC 9.8 and GHC 9.10

DigitalBrains1:
This PR includes the commits of PR #110 (which I initially had not noticed when I opened this PR).

Updates the `circuit-notation` plugin to current master.

`Protocols.Internal` actually contains quite some orphan instances regardless of `clash-prelude` version now.

Rowan or Tim, could you verify I did the right thing in the _Fix partial function warnings in PacketStream_ commit? By the time I got to `makePropPacketArbiter` I was frankly spent and my error message is pretty non-helpful.

Funny thing: `head` has a warning but `last` doesn't. So we could just rewrite every use of `head` to `last . reverse`. For some reason I didn't.

### DIFF
--- a/.ci/apply_settings.sh
+++ b/.ci/apply_settings.sh
@@ -1,0 +1,8 @@
+#!/bin/bash
+
+if [[ "$check_haddock" != @(True|False) ]]; then
+  echo "check_haddock: Expected True or False, got \"$check_haddock\"" >&2
+  exit 1
+fi
+sed <.ci/cabal.project.local.in >cabal.project.local "
+    s/__CHECK_HADDOCK__/$check_haddock/"
--- a/.ci/build_docs.sh
+++ b/.ci/build_docs.sh
@@ -1,21 +1,8 @@
 #!/bin/bash
 set -xeou pipefail
 
-# Build dependencies first, so they don't end up in logs
-cabal v2-build \
-  --constraint=clash-prelude==$clash_version \
-  --enable-documentation \
-  --allow-newer=circuit-notation:ghc \
-  clash-protocols
-
-# circuit-notation currently _compiles on 8.10, but isn't usable. The only
-# other GHC version it supports is 8.6.5, but this GHC bundles a Haddock that
-# cannot generate documentation for clash-prelude. Hence, we build docs with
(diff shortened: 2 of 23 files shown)


rowanG077 (approved):
Seems very uncontroversial to me. Once you got CI working it can be merged from my perspective.

bgamari:
I suppose this supercedes #110?

DigitalBrains1:
> I suppose this supercedes #110?

Oh! I've hardly ever worked on this repository and completely forgot to look at existing PR's.

My suggestion is to incorporate those commits into this PR, do you agree? I can also do it the other way around but that would require pushing to your fork; it seems simpler to use this branch.

bgamari:
> > I suppose this supercedes #110?
> 
> Oh! I've hardly ever worked on this repository and completely forgot to look at existing PR's.
> 
> My suggestion is to incorporate those commits into this PR, do you agree? I can also do it the other way around but that would require pushing to your fork; it seems simpler to use this branch.

Yes, feel free the take my changes. I just wanted to make sure we avoided duplication of work where possible.

.ci/apply_settings.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+
+if [[ "$check_haddock" != @(True|False) ]]; then
+  echo "check_haddock: Expected True or False, got \"$check_haddock\"" >&2
+  exit 1
+fi
+sed <.ci/cabal.project.local.in >cabal.project.local "
+    s/__CHECK_HADDOCK__/$check_haddock/"
```
martijnbastiaan, line 8: Why not `sed -i`?

martijnbastiaan (approved):
Seems uncontroversial :)

.ci/apply_settings.sh
DigitalBrains1: But then you'd still have to copy the file, right?

If I try to use `--in-place`, I end up with something like:

```
sed .ci/cabal.project.local.in -i -e "
    s/__CHECK_HADDOCK__/$check_haddock/"
cp .ci/cabal.project.local.in cabal.project.local
```

PR #132, opened 2025-01-28
Small improvements to CI

DigitalBrains1:
I noticed a few points where CI could easily be improved a bit. See the individual commit messages for details.

I also bumped the Stack version in `stack.yaml` to the latest LTS.

PR #133, opened 2025-01-29
Support Clash current master

DigitalBrains1:
The type of `blockRamU` has changed since Clash 1.8.

The `clash-cores` package explicitly targets both Clash 1.8 and Clash master. Since we want to add `clash-protocols` as a dependency to `clash-cores`, we need `clash-protocols` to support Clash master.

### DIFF
--- a/clash-protocols/src/Protocols/Df.hs
+++ b/clash-protocols/src/Protocols/Df.hs
@@ -902,6 +902,28 @@ registerBwd =
     iDatX1 = C.regEn (C.errorX "registerBwd") oAck iDatX0
     oDat = toData <$> valid <*> (C.mux oAck iDatX0 iDatX1)
 
+-- Fourmolu only allows CPP conditions on complete top-level definitions. This
+-- function is not exported.
+blockRamUNoClear ::
+  forall n dom a addr.
+  ( HasCallStack
+  , C.HiddenClockResetEnable dom
+  , C.NFDataX a
+  , Enum addr
+  , C.NFDataX addr
+  , 1 <= n
+  ) =>
+  C.SNat n ->
+  Signal dom addr ->
+  Signal dom (Maybe (addr, a)) ->
+  Signal dom a
+#if MIN_VERSION_clash_prelude(1,9,0)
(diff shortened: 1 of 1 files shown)


DigitalBrains1:
The CI failure in GHC 9.2.8 (Haddock) has nothing to do with this PR. I don't know what broke, but I don't even care much because PR #132 makes it so we only test Haddock on GHC 9.6, neatly getting rid of this failure in the process.

clash-protocols/src/Protocols/Df.hs
```diff
@@ -898,6 +898,26 @@ registerBwd =
     iDatX1 = C.regEn (C.errorX "registerBwd") oAck iDatX0
     oDat = toData <$> valid <*> (C.mux oAck iDatX0 iDatX1)
 
+blockRamUNoClear ::
```
martijnbastiaan, line 901: Perhaps copy+paste the documentation of `blockRamU` and explain that it removes the clear logic.

martijnbastiaan (approved):

clash-protocols/src/Protocols/Df.hs
DigitalBrains1: Luckily I don't actually _export_ this identifier, I wouldn't want to do that. The only reason I don't use a local definition is that Fourmolu would get mad at me. The only way to `#if` an expression or definition with Fourmolu is to create a top-level binder for the expression like I do here. You can't do `#if` in the middle of an expression and you can't do `#if` with a local definition (`let` or `where` block, although frankly I've never tried a `let`, it seems unlikely it would work).

So I can add some explanation, but it would not affect the generated Haddock or the export list of this module.

martijnbastiaan: Ahh, ok. Yeah a quick note saying that this is a Fourmolu workaround is fine. (Or disable Fourmolu locally if it's too much of a hassle?)

DigitalBrains1:
Oof I blundered. I forgot to do the same for test cases. But CI doesn't mind because it only tests with Clash 1.8, not with Clash master.

That was sloppy of me.

PR #136, opened 2025-02-25
Fix use of deprecated OPTIONS pragma

DigitalBrains1:
Per [the documentation](https://downloads.haskell.org/ghc/latest/docs/users_guide/exts/pragmas.html#options-ghc-pragma):

> Previous versions of GHC accepted `OPTIONS` rather than `OPTIONS_GHC`, but that is now deprecated.

Additionally, the Haddock module description seemed to refer to the `-fplugin` option rather than actually describing the module, so the comment is moved and made a regular comment instead of a Haddock block. For consistency with the rest of the code base, line comments are used instead of block comments.

### DIFF
--- a/clash-protocols/tests/Tests/Protocols/Plugin.hs
+++ b/clash-protocols/tests/Tests/Protocols/Plugin.hs
@@ -1,13 +1,11 @@
 {-# LANGUAGE BlockArguments #-}
-
-{-# OPTIONS -fplugin=Protocols.Plugin #-}
+-- This /must/ be enabled in order for the plugin to do its work. You might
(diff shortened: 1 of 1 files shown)


DigitalBrains1:
Ah, that makes sense! When reformatting everything for Fourmolu, the comment [got moved](https://github.com/clash-lang/clash-protocols/commit/0fe03b12f680aabd3cdf074c1af4d6544564a044#diff-5f1d342333d82bf845e920d095ea8905a68fe6c01e504ca4ef7a7d2aabb89e96). Since it had been inadvertently marked as a Haddock block, it was moved into the proper place for that even though it wasn't supposed to be Haddock.

PR #138, opened 2025-04-24
Fix Df.roundRobinCollect adherence to the Df proto

### LINKED ISSUE #137: `roundrobinCollect Parallel` violates the `Df` protocol
The `Parallel` mode of `Protocols.Df.roundrobinCollect` advertises that it should:

> Check all components in parallel. Biased towards the /last/ Df channel.

Recall that the `Df` protocol is bound by the invariant that a value presented downstream must not change until an acknowledgement is presented by the consumer.

In light of this, it is clear that implementing these semantics requires state. Consider, for instance, an application of `roundrobinCollect Parallel` with two inputs. At `t=0` the first input  may present `roundrobinCollect` with a value `a`; this value will be propagated downstream. Then at `t=1` the second input may present `roundrobinCollect` with `b`. Under the above invariant , `roundrobinCollect` must not change its output until the input `a` has been accepted by its downstream consumer; this requires state.

However, at the moment `roundrobinCollect Parallel` is implemented thusly:
```haskell
roundrobinCollect Parallel =
  Circuit (B.first C.unbundle . C.unbundle . fmap go . C.bundle . B.first C.bundle)
 where
  go (dats0, ack) = (acks, dat1)
   where
    nacks = C.repeat (Ack False)
    acks = Maybe.fromMaybe nacks ((\i -> C.replace i ack nacks) <$> iM)
    dat1 = Maybe.fromMaybe NoData dat0
    (iM, dat0) = Data.List.NonEmpty.unzip dats1
    dats1 = C.fold @(n C.- 1) (<|>) (C.zipWith goDat C.indicesI dats0)

    goDat i dat
      | Maybe.isJust (dataToMaybe dat) = Just (i, dat)
      | otherwise = Nothing
```

This clearly does not abide the `Df` protocol.

rowanG077:
It used to switch the sources even when the current output was not acked.

Fixes #137

### DIFF
--- a/clash-protocols/src/Protocols/Df.hs
+++ b/clash-protocols/src/Protocols/Df.hs
@@ -88,6 +88,7 @@ module Protocols.Df (
   noData,
   fromData,
   toData,
+  dataDefault,
 ) where
 
 -- base
@@ -114,11 +115,6 @@ import Prelude hiding (
 import qualified Data.Bifunctor as B
 import Data.Bool (bool)
 import qualified Data.Coerce as Coerce
-#if MIN_VERSION_base(4,19,0)
-import qualified Data.Functor as Functor (unzip)
-#else
-import qualified Data.List.NonEmpty as Functor (unzip)
-#endif
 import Data.Kind (Type)
 import Data.List ((\\))
 import qualified Data.Maybe as Maybe
@@ -223,6 +219,14 @@ fromData (Data a) = a
 toData :: Bool -> a -> Data a
 toData False _ = NoData
 toData True a = Data a
+
+{- |
+  If the t'Data' is v'NoData', it returns the given value; otherwise,
+  it returns the value contained in the v'Data'.
+-}
+dataDefault :: a -> Data a -> a
+dataDefault a NoData = a
+dataDefault _ (Data a) = a
 
 instance (C.KnownDomain dom, C.NFDataX a, C.ShowX a, Show a) => Simulate (Df dom a) where
   type SimulateFwdType (Df dom a) = [Data a]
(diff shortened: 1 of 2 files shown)


clash-protocols/src/Protocols/Df.hs
```diff
@@ -224,6 +219,14 @@ toData :: Bool -> a -> Data a
 toData False _ = NoData
 toData True a = Data a
 
+{- |
+  If the Data is `NoData`, it returns the gi value; otherwise,
+  it returns the value contained in the Data.
```
martijnbastiaan, line 224:
```suggestion
  If the "Data" is 'NoData', it returns the given value; otherwise,
  it returns the value contained in the 'Data'.
```

" -> type
' -> term

AFAIK

martijnbastiaan (approved):
Could you add an issue saying the protocol checkers should check that data doesn't change while the "valid bit" is set?

clash-protocols/src/Protocols/Df.hs
DigitalBrains1:
```suggestion
  If the t'Data' is 'NoData', it returns the given value; otherwise,
  it returns the value contained in the v'Data'.
```

Don't you mean
```
t'Type'
v'Term'
```
? https://haskell-haddock.readthedocs.io/latest/markup.html#hyperlinked-identifiers

Note you need GHC 9.0 or newer to get a Haddock that understands it, but this package already requires that.

DigitalBrains1:
> Could you add an issue saying the protocol checkers should check that data doesn't change while the "valid bit" is set?

Relevant:
https://github.com/clash-lang/clash-protocols/blob/febb041c32fa678e7ddb7d0e976516a5c56b03fc/README.md?plain=1#L84

clash-protocols/src/Protocols/Df.hs
```diff
@@ -224,6 +220,14 @@ toData :: Bool -> a -> Data a
 toData False _ = NoData
 toData True a = Data a
 
+{- |
+  If the t'Data' is v'NoData', it returns the given value; otherwise,
+  it returns the value contained in the v'Data'.
+-}
+dataDefault :: a -> Data a -> a
+dataDefault a NoData = a
+dataDefault _ (Data a) = a
```
bgamari, line 229: `fromData` is arguably a more idiomatic name for this (c.f. `fromMaybe`).

martijnbastiaan:
You have merge rights, right @rowanG077?

clash-protocols/src/Protocols/Df.hs
rowanG077: Agreed but that function exists already as:

```
-- | Extract value from `Data`, Bottom on `NoData`.
fromData :: (HasCallStack, C.NFDataX a) => Data a -> a
fromData NoData = C.deepErrorX "fromData: NoData"
fromData (Data a) = a
```

PR #141, opened 2025-05-07
AXI: Add `BitPack` instance to all channel signals

hiddemoll:
To be honest, I'm not sure why these datatypes did not already have a `BitPack` instance.

### DIFF
--- a/clash-protocols/src/Protocols/Axi4/Common.hs
+++ b/clash-protocols/src/Protocols/Axi4/Common.hs
@@ -1,8 +1,4 @@
-{-# LANGUAGE GADTs #-}
 {-# LANGUAGE TypeFamilyDependencies #-}
-{-# LANGUAGE UndecidableInstances #-}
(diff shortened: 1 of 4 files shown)


DigitalBrains1:
Could you add a bit to the commit that switches strict to lazy tuples on why this is okay?

Also, on a larger timetable, it seems to me we should make it possible to derive `BitPack` for types involving strict tuples, since people might want to use those with good reason.

martijnbastiaan:
> Could you add a bit to the commit that switches strict to lazy tuples on why this is okay?

I'll jump in: I started out `clash-protocols` with many strict data types out of a fear of space leaks. But this was never really well defended in the first place, so IMO we should just do what the rest of the Haskell ecosystem does by default: laziness.

hiddemoll:
I've added Martijns comment to the commit body. Do you agree on the formatting of that commit body?

PR #143, opened 2025-05-09
Hotfix: Fallout from Data to Maybe migration

rowanG077:
made me mess up a check

### DIFF
--- a/clash-protocols/src/Protocols/Df.hs
+++ b/clash-protocols/src/Protocols/Df.hs
@@ -798,9 +798,9 @@ roundrobinCollect Parallel =
     fwd = fwds C.!! i
 
     nextIm =
-      if Maybe.isJust fwd && ack
-        then im
-        else nextSrc
+      if Maybe.isNothing fwd || ack
+        then Nothing
+        else im
 
 -- | Place register on /forward/ part of a circuit. This adds combinational delay on the /backward/ path.
 registerFwd ::


rowanG077:
And also now looking at it again it used to be suboptimal in terms of throughput.

Consider this code:

```Haskell
  go im (fwds, bwd@(Ack ack)) = (nextIm, (bwds, fwd))
   where
    nextSrc = C.fold @(n C.- 1) (<|>) (C.zipWith (<$) C.indicesI fwds)
    i = Maybe.fromMaybe (Maybe.fromMaybe maxBound nextSrc) im

    bwds = C.replace i bwd (C.repeat (Ack False))
    fwd = fwds C.!! i

    nextIm =
      if Maybe.isNothing fwd || ack
        then nextSrc
        else im
```

if `im` is `Nothing` and there some forward sending something. then `nextSrc` will be `Just i`. So that `i` is selected to be connected. Great. But then assume that it is instantly acked. `nextIm` will become `Just i` so in the next cycle the component will be forced to connect the previous selected forward. But that forward may no longer be producing anything. Only then in the next cycle it can potentially switch to a new forward. So it wastes a cycle when switching. 

But explicitly setting `nextIm` to `Nothing` when the currently selected forward has been acked fixes this.

martijnbastiaan:
So this is purely a performance fix? If no, why didn't the tests catch it?

rowanG077:
It didn't catch it for the same reason for the same reason the original issue existed. `propWithModel` doesn't test for it.

martijnbastiaan (approved):

PR #148, opened 2025-06-06
Rework Nix system

jaschutte:
Removed all 'old-style' Nix infrastructure in favor of nix flakes.

This makes all the Nix logic easier to understand whilst also making it easier to include this within other projects.

The only downside is that you can no longer use old nix syntax, but considering clash-compiler doesn't support that, I don't see why clash-protocols should.


martijnbastiaan:
Stupid question, but why does it build `clash-protocols` in a `nix shell`? I thought that was supposed to give you an environment where you could start developing _for_ the repo you're in.

jaschutte:
Not at all! It's actually misleading naming, it doesn't build clash-protocols itself. In `shellFor` the packages attribute is a list of packages you *plan* to build. As I understand, you have to mention the package you want to build within the packages attribute so Nix can know which dependencies it should expose to cabal. So it will build all *dependencies* of clash-protocols (so it can give those to cabal), but not clash-protocols itself.

rowanG077:
`mkShell` does have a similar thing. You can add packages to the `inputsFrom` attribute. Which essentially does what you describe as well.

rowanG077:
That patch should no longer be needed since this PR moves to GHC 9.10 and this has long since been upstreamed to GHC.

DigitalBrains1 (changes requested):
Per Rowan's comment, the patch for aarch64 should remain

[edit] Or GHC upgraded from current 9.10.1 to 9.10.2 [/edit]

DigitalBrains1:
> I do wonder why the patches are applied here though instead of directly at clash-compiler, surely if it's a bug with GHC it would happen to all clash projects right and not just clash-protocols? So why are the patches being applied only here?

Excellent question! The answer might be that Rowan [critically needed it](https://github.com/clash-lang/clash-protocols/pull/88) to continue what he was working on. But if you can actually make this part of the Nix flake in the `clash-compiler` repo and have it affect downstream uses like `clash-protocols`, that sounds like a good improvement! I requested it to be retained here to not introduce an actual regression, it's not that I oppose better solutions :-).

DigitalBrains1 (approved):

DigitalBrains1:
Rowan, I agree with James that it'd be best if someone actually tested whether the `aarch64` patching Nix code truly accomplishes its task. Could you test it on an `aarch64-linux` system? Alternatively, I suppose I could use QEMU to spin up a VM with an emulated processor, if you currently have no time to spend on this.

DigitalBrains1:
I think this should be squash-merged, with the commit message of the first commit as the final commit message. You could make this more explicit by squashing the commits before merge (but we'll still use a squash merge as there is a difference between a squash merge and a normal merge of a single commit, and we like to use the former).

rowanG077:
Currently the patch is not applied because the flake refers to the upstream clash package `clash-compiler.packages` in the overlay which doesn't contain the patched GHC. But worse. the upstream clash-compiler flake has no linux-aarch64 support. For now I would say just remove the patch. I will take care of adding it upstream to clash-compiler flake when I want to do development soon.

rowanG077 (approved):

jaschutte:
Thank you! I removed the patch for now.

PR #159, opened 2025-07-07
Add `truncateAbortedPackets`

lmbollen:
This component enables the user to truncate aborted packets.
For some applications it may not be useful to keep processing a packet for which we know that it holds no value, this component then allows you to truncate it to free up the pipeline.


martijnbastiaan (approved):
LGTM, modulo 1 comment. Also agree with @rowanG077's comment.

rowanG077 (approved):
Minor style things.

PR #162, opened 2025-07-24
Fix broken flake for newer versions of Clash

jaschutte:
Newer versions of Clash (ever since 583b32e) broke the clash-protocols flake due to clash-protocols requiring tasty < 1.5 and Clash tasty >= 1.5 and having other dependency issues. This PR fixes those dependency issues by refactoring the flake. It also raised the bound of tasty from <1.5 to <1.6.

Overall the flake is a bit 'cleaner'. Rather than manually applying the same fixes clash-compiler applies, it reuses the same set of packages from clash-compiler now.

### DIFF
--- a/clash-protocols/clash-protocols.cabal
+++ b/clash-protocols/clash-protocols.cabal
@@ -110,7 +110,7 @@ library
     , template-haskell
 
       -- To be removed; we need 'Test.Tasty.Hedgehog.Extra' to fix upstream issues
-    , tasty >= 1.2 && < 1.5
+    , tasty >= 1.2 && < 1.6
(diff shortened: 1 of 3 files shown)


DigitalBrains1:
I don't know much about Nix, but wouldn't it make more sense to just change the .cabal file instead of patching it inside of the Nix flow? Is the bound just too restrictive or is there something special about Nix that means that even though it normally needs an older `tasty` it does work specifically for Nix?

DigitalBrains1:
Also please note that in `clash-compiler`, we use a custom Tasty both in `cabal.project` and `stack.yaml`, not the default Tasty.

PR #176, opened 2025-09-29
Simplify CI caching logic

martijnbastiaan:
Previous implementation contained a bunch of references to non-existing variables.


martijnbastiaan:
Also, I wonder if there is a way to have one `clash-lang` repo where we can scrutinize CI scripts. Every repo seems to be doing it a bit different.

PR #178, opened 2025-09-30
Move `applyC` to an appropriate location and ensure it is exported

rslawson:
Forgot to do this in #174.


martijnbastiaan (approved):
Well we do export `.Internal`, it's just a very strong signal to users that they shouldn't use it. But exporting seems good. 

I'm happy to merge, but I'd like a second opinion. @rowanG077 What do you think? I like `applyC`, because some implementations become much cleaner, see for example:

https://github.com/clash-lang/clash-protocols/pull/174/files

PR #196, opened 2026-04-22
Upstream `clash-protocols-memmaps` dependencies to experimental package

hiddemoll:
**_What (what did you do)_**
Create a new package `clash-protocols-experimental` where we can test-drive functionality before releasing it in `clash-protocols` when the API is stable.

**_Why (context, issues, etc.)_**
@hydrolarus is in the process for moving `clash-protocols-memmap` and `clash-bitpackc` to its own repo. These packages depend on some additions to `clash-protocols` which we collected in `bittide-extra`. The specific dependencies of `clash-protocols-memmap` are:
- `Protocols.Wishbone.Extra (delayWishbone)`
- `Protocols.ReqResp`
These in turn depend on `Protocols.BiDf` and some smaller additions to `Df`.

**_Dear reviewer (anything you'd like the reviewer to pay close attention to?)_**
This is a big PR adding functionality for which we'd normally open separate PRs. I've broken the work down into reviewable chunks, so I recommend reviewing per commit. Pretty much everything is a direct copy-paste from `bittide-hardware`.

Some genenal points of attention:
- I've copy-pasted the `clash-protocols` package, removing unnecessary parts. I could have made mistakes in reducing the .cabal file.
- `Protocols.FunctorC`: Does not have a unittest, but IMO it doesn't need one as it is so small and simple.
- `Protocols.Df.Extra`: I only copied the functions needed for `clash-protocols-memmap` (`skid`, `bypassFifo` and `stallNext`), including their unittests.
- `Protocols.BiDf`: Copied everything except the `fromBlockRam(WithMask)` functions.
- `Protocols.ReqResp`: Copied everything except the `fromBlockRam(WithMask)` functions.
- `Protocols.Wishbone.Extra`: Copied `delayWishbone`, added new unittest (with help from Claude).

**_AI disclaimer (heads-up for more than inline autocomplete)_**
I used Claude to write unittest for `delayWishbone`. It still largly follows the same structure as in `bittide`, but without using `wbStorage` which depends on `clash-protocols-memmap`.

# TODO
_~~Cross-out~~ any that do not apply_
- [x] Write (regression) test
  -> except for `fmapC`
- [x] Link to existing issue
  -> Linked to the orginal PRs on `clash-protocols` or `bittide-hardware` where possible


clash-protocols/src/Protocols/Experimental/Df/Extra.hs
```diff
@@ -0,0 +1,180 @@
...
+  deriving (Generic, NFDataX)
+
+{- | A skid buffer that can store a single element and will acknowledge data if it can
+store it. Also indicates when it is ready to accept data. In practice, the ready signal is
+the exact same as the ack signal, but 'Df' disallows interpreting the ack signal as a
+this-component-is-ready-to-accept-data signal.
+
+This skid buffer is implemented in terms of 'moore', so it cuts up combinational paths.
+-}
+skid ::
```
rowanG077: Please document that this cannot run at full throughput and inserts bubbles every other cycles under full input throughput. I don't think this is a skid buffer because of this. So a better name would maybe be `singleEntryBuffer`?

clash-protocols/tests/Tests/Protocols/BiDf.hs
```diff
@@ -0,0 +1,139 @@
...
+    (\_ _ _ -> id)
+    (exposeClockResetEnable impl)
+    prop
+ where
+  simConfig = def
+  eOpts = defExpectOptions{PH.eoStopAfterEmpty = Just 300}
+  gen :: Gen (Vec 3 [Int])
+  gen = genVec (Gen.list (Range.linear 0 10) smallInt)
+  -- Since stalling can change the order of the samples, we only check if they are all present
+  prop expected sampled = L.sort (L.concat (toList sampled)) H.=== L.sort (L.concat (toList expected))
```
rowanG077: This does not catch the key property that the responses go to the right requests. Could you add a tag to each request with it's index and check each source only receives responses meant for itself?

clash-protocols/src/Protocols/Experimental/BiDf.hs
```diff
@@ -0,0 +1,167 @@
...
+
+-- | Map both requests and responses.
+bimap ::
+  (req -> req') ->
+  (resp -> resp') ->
+  Circuit (BiDf dom req resp') (BiDf dom req' resp)
+bimap f g = map (Df.map f) (Df.map g)
+
+-- | Merge a number of 'BiDf's, preferring requests from the last channel.
+fanin ::
```
rowanG077, line 122: Documentation is wrong, it prefers the first channel not the last.

rowanG077 (changes requested):
Small changes

clash-protocols/src/Protocols/Experimental/BiDf.hs
rowanG077: This function introduces bubbles too. That's really not what I expect from a `fanin`.

lmbollen: We should really get some infrastructure to programmatically test whether components can run at full throughput. It seems to me like this should not be very difficult to write up in a hedgehog test.

clash-protocols/src/Protocols/Experimental/Df/Extra.hs
hiddemoll: Nice find. I've updated the name to `singleEntryBuffer` (rip nice short name of `skid`), and updated the docs to reflect it inserts bubbles.

clash-protocols/tests/Tests/Protocols/BiDf.hs
hiddemoll: You have a good eye for things that could be wrong, because this was actually wrong! Not because of the implementation of `fanin`, but I  (with help from Claude) found a bug in `Df.roundrobinCollect`. I fixed that, and let Claude add a regression test.

clash-protocols/tests/Tests/Protocols/Df.hs
```diff
@@ -360,6 +360,46 @@ case_roundrobinCollectParallel = do
       C.withClockResetEnable @C.System C.clockGen noReset C.enableGen $
         Df.roundrobinCollect @3 Df.Parallel
 
+{- | Asserts that roundrobinCollect with Parallel mode commits to a source once
```
rowanG077, line 363: If you use unique payloads this check reduces to checking if the output is constant until acked (baseline DF property). I think that's much simpler.

clash-protocols/src/Protocols/Experimental/BiDf.hs
hiddemoll: I think the priority should be to get this PR merged as soon as possible. Rewriting `BiDf.fanin` to _not_ introduce bubbles is out-of-scope IMO. Add performance testing infrastructure is definetly out-of-scope for this PR. 

Instead, I updated the docs of `BiDf.fanin` so it says it introduces bubbles. Currently `BiDf.fanin` is not used anywhere, so I think this falls within the mindset of "this ~package~ module is experimental".

clash-protocols/tests/Tests/Protocols/Df.hs
hiddemoll: ~Currently `roundrobinCollect Parallel` can switch channels mid-transaction, which this new test catches.~
I see you edited your comment, so this comment of mine is not related

clash-protocols/src/Protocols/Experimental/BiDf.hs
rowanG077: It's fine by me to merge it as is. Adding a new test fixture is indeed out of scope I would say. Fixing the bubbles is debatable and I wouldn't accept it in a non-experimental module. But for now if it is documented with a fat warning I'm oke with it.

clash-protocols/tests/Tests/Protocols/Df.hs
hiddemoll: Added unique payloads. Is this the simplification you had in mind?

rowanG077: Yeah sorry, you were really fast with reading it :p. Yes this is what I had in mind.

rowanG077 (approved):

PR #197, opened 2026-04-22
Add default implementation for `Fwd` and `Bwd`

martijnbastiaan:
The more I use `Circuit`s, the more I use the following pattern to "lift" types containing signals to circuit notation:

```haskell
data Input dom = Input
  { coreReset :: Reset dom
  , subCoreReset :: Reset dom
  , wibbles :: Signal dom (Unsigned 64)
  , wobbles :: Signal dom (Unsigned 64)
  }
```

I mostly do that to get away from positional arguments / boolean(like) blindness. Lifting that to `Circuit` then looks like:

```haskell
instance Protocol (Input dom) where
  type Fwd (Input dom) = Input dom
  type Bwd (Input dom) = ()

instance Protocol (Output dom) where
  type Fwd (Output dom) = Output dom
  type Bwd (Output dom) = ()
```

..at which point I can conveniently use it using `Fwd` and friends. To me this sounds like a sensible default implementation.

### DIFF
--- a/clash-protocols-base/src/Protocols/Plugin/Types.hs
+++ b/clash-protocols-base/src/Protocols/Plugin/Types.hs
@@ -19,9 +19,13 @@ class Protocol a where
   -- existence of 'Fwd'.
   type Fwd (a :: Type)
 
+  type Fwd a = a
+
   -- | Receiver to sender type family. See 'Circuit' for an explanation on the
   -- existence of 'Bwd'.
   type Bwd (a :: Type)
+
+  type Bwd a = ()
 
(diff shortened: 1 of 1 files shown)


lmbollen (approved):

martijnbastiaan:
@lmbollen and I discussed it in the office. We were thinking another sensible default implementation would be `Fwd a = Signal dom a`, however, this is of course impossible because it conjures up the `dom` out of nowhere. We therefore think this default implementation is almost as unopinionated as you can get.

rowanG077 (approved):
I don't see a problem with this.

PR #198, opened 2026-04-23
Prepare release

rowanG077:
Fixes for things I found while looking around the package to make it releasable. See commits for details. The most important is the move towards clash 1.10 and ghc 9.12.4 support. 

The biggest code movement is that the `Avalon` and `Axi` protocols are moved to experimental as I think both of them will change in the future. There are also a lot of TODOs in them. Were these also the ones you were thinking of when you mentioned this to me @lmbollen? I considered an experimental package but a module for now is easier imo.


clash-protocols/clash-protocols.cabal
```diff
@@ -1,13 +1,14 @@
...
+name: clash-protocols
+synopsis: a battery-included library for (dataflow) protocols
+homepage: https://github.com/clash-lang/clash-protocols
+bug-reports: https://github.com/clash-lang/clash-protocols/issues
+version: 0.1
+category: Hardware
+license: BSD-2-Clause
+license-file: LICENSE
+author: Martijn Bastiaan, QBayLogic B.V.
+maintainer: Martijn Bastiaan <martijn@hmbastiaan.nl>
```
martijnbastiaan:
```suggestion
maintainer: QBayLogic B.V. <devops@qbaylogic.com>
```

clash-protocols/clash-protocols.cabal
```diff
@@ -39,95 +47,92 @@ common common-options
...
+    -- `-fexpose-all-unfoldings` is required because clash needs access to the
+    -- source code in compiled modules
+    -Wall
+    -Wcompat
+    -fplugin
+    GHC.TypeLits.Extra.Solver
+    -fplugin
+    GHC.TypeLits.Normalise
+    -fplugin
+    GHC.TypeLits.KnownNat.Solver
```
martijnbastiaan:
```suggestion
    -fplugin=GHC.TypeLits.KnownNat.Solver
```

martijnbastiaan (approved):
I've got no blockers, approve :). Questions:

What makes `Wishbone` less experimental than the others? 

Would you consider moving the `Simulate` class to experiment too? They were written in a time before I was comfortable with `circuit-notation` and the test infrastructure turned out very ad-hoc. There is definitely parts I like, but it has also grown into a quite complex web with a lot of assumptions about protocols. I hope to move us to a situation where it would be much more (excuse my french) UVM-like: per-protocol drivers, consumers, dashboards, etc.

clash-protocols/clash-protocols.cabal
rowanG077: This is auto-formatted by `cabal-gild` I don't like it either, but I also don't want to break the formatting.

rowanG077:
> What makes Wishbone less experimental than the others?

From reading the code wishbone standard seems more complete than the other two. I was on the fence to put it into experimental. I guess you use this in bittide? So you have way more experience with it than me. So would you put it into experimental as well?

> Would you consider moving the Simulate class to experiment too? They were written in a time before I was comfortable with circuit-notation and the test infrastructure turned out very ad-hoc. There is definitely parts I like, but it has also grown into a quite complex web with a lot of assumptions about protocols. I hope to move us to a situation where it would be much more (excuse my french) UVM-like: per-protocol drivers, consumers, dashboards, etc.

Yes, I agree.

clash-protocols/clash-protocols.cabal
martijnbastiaan: Notice the `=` :)

rowanG077: Ah that makes a diffference, interesting.

martijnbastiaan:
> So would you put it into experimental as well?

I'm also on the fence. The current definition serves us well in bittide and there are no open complaints, so marking it stable is okay by me. I was just curious to hear your thoughts.

martijnbastiaan:
By the by, I don't think you're going to get a reply from Lucas before Monday/Tuesday.

rowanG077:
Then for now I will do the conservative thing and move wishbone as well to experimental. Moving it back to non-experimental is something we can do at any time but the other way around is hard :).

lmbollen:
I'm working to day so I'm available.

Yes I think it's a good idea to move Axi and Avalon.

Regarding Wishbone, it seems as a protocol to be simple enough to have all the fields all the type. The only "complaint" is that the current implementation is not very haskell-esque. (No use of ADTs for example), but maybe that's something we should just reserve for another time.

martijnbastiaan:
> I'm working to day so I'm available.

Right, I forgot :facepalm:.

> The only "complaint" is that the current implementation is not very haskell-esque.

Yeah. We talked before about protocols maybe coming in two flavors: one you could use at the borders of your design (top entities or when talking to IP), and one to use internally.

PR #201, opened 2026-04-29
Fix `stallStandard`, the implementation did not work at all.

lmbollen:
When I tried to apply the previous implementation of `stallStandard` I could not get it working and discovered it is not  used in either `clash-protocols` or `bittide-hardware`.

So I simplified the implementation such that the only thing `stallStandard` does is block the outgoing strobe (and or busCycle) signals to artificially stall the manager.

I managed to use the function on bittide, currently looking into making test infra for it.

### DIFF
--- a/clash-protocols/src/Protocols/Experimental/Wishbone/Standard/Hedgehog.hs
+++ b/clash-protocols/src/Protocols/Experimental/Wishbone/Standard/Hedgehog.hs
@@ -62,7 +62,6 @@ import Clash.Hedgehog.Sized.Unsigned (genUnsigned)
 import Clash.Prelude as C hiding (cycle, indices, not, sample, (&&), (||))
 import Clash.Signal.Internal (Signal ((:-)))
 import Control.DeepSeq (NFData)
-import Data.Bifunctor qualified as B
 import Data.List.Extra
 import Data.String.Interpolate (i)
 import GHC.Stack (HasCallStack)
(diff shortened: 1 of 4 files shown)


clash-protocols/tests/Tests/Protocols/Wishbone.hs
```diff
@@ -287,11 +296,212 @@ prop_specViolation_lenient = property $ do
...
+            monitoredCircuit = circuit $ \wb -> do
+              (wb1, preSig) <- circuitMonitor -< wb
+              wb2 <- stallStandard earlyBusCycle memStalls -< wb1
+              (wb3, postSig) <- circuitMonitor -< wb2
+              memoryWb (blockRam (C.replicate d256 0)) -< wb3
+              idC -< (preSig, postSig)
+           in
+            driveStandard eOpts reqPairs |> monitoredCircuit
+
+    verifyStalls (n : stalls) trace =
```
rowanG077, line 468: I think it's possible that trace has fewer cycles than expected and then it should report a failure. Currently this is silently accepted.

clash-protocols/src/Protocols/Experimental/Wishbone/Standard/Hedgehog.hs
```diff
@@ -303,83 +302,54 @@ stallStandard ::
...
     (Wishbone dom 'Standard addressBits dataBytes)
     (Wishbone dom 'Standard addressBits dataBytes)
-stallStandard stallsPerCycle =
-  Circuit $
-    B.second (emptyWishboneM2S :-)
-      . uncurry (go stallsPerCycle Nothing)
+stallStandard earlyBusCycle stalls0 =
+  Circuit goS
  where
+  goS (m2s, s2m) = (s2m, go stalls0 m2s s2m)
```
rowanG077: `s2m` is passed through while stalling. Shouldn't that be forced to non-acked?

rowanG077 (changes requested):

rowanG077 (approved):

PR #203, opened 2026-05-27
Make all Nix (shells) buildable

martijnbastiaan:
This doesn't add it to CI yet, I'll rebase https://github.com/clash-lang/clash-protocols/pull/167 on top of this.

Two opinionated fixes:

* I'm copying CHANGELOG.md in the Nix expression. This hack should disappear over time, once we drop support for older GHCs/Cabals.
* I couldn't get `cabal-gild` to work for the GHC 9.6.7. I thought about overriding it to another version, but that would mean different formatters in different shells. Instead I just disabled it for the old GHC.

### DIFF
--- a/clash-protocols-base/clash-protocols-base.cabal
+++ b/clash-protocols-base/clash-protocols-base.cabal
@@ -57,7 +57,7 @@ common common-options
 
   default-language: GHC2021
   build-depends:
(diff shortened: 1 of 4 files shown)


rowanG077 (approved):
That is a quite a pile of fixups. It is great the other GHC are working now!

